PR #52, opened 2022-11-11
refactor: adding qwik-ui core to the documentation website

gilf:
Adding qwik-ui core to the documentation website


packages/core/project.json
```diff
@@ -32,7 +32,7 @@
       "outputs": ["{options.outputPath}"],
       "options": {
         "uiFramework": "@storybook/html",
-        "outputPath": "dist/storybook/core",
+        "outputPath": "packages/core/dist",
```
shairez, line 35: This should be inside of the `packages`

nx convention is to use `dist/[package-name]/ ... `

gilf: reverted

shairez: Thanks!

PR #89, opened 2023-01-24
Tooltip flickering

nnelgxorz:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Screenshots/Demo

# Checklist:

### DIFF
--- a/packages/headless/src/components/tooltip/tooltip.tsx
+++ b/packages/headless/src/components/tooltip/tooltip.tsx
@@ -1,6 +1,7 @@
 import {
   $,
   component$,
+  QRL,
   Slot,
   useClientEffect$,
   useId,
@@ -8,7 +9,7 @@ import {
   useSignal,
   useStylesScoped$,
 } from '@builder.io/qwik';
-import { computePosition, flip, type ComputePositionConfig } from '@floating-ui/dom';
+import { autoUpdate, computePosition, flip, type ComputePositionConfig } from '@floating-ui/dom';
 import styles from './tooltip.css?inline';
 
 export interface TooltipProps {
@@ -48,7 +49,6 @@ export const Tooltip = component$(
             top: `${y}px`
           })
         });
-        stateSignal.value = 'positioned';
       }
     });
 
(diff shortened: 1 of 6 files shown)


packages/website/src/routes/daisy/tooltip/index.tsx
```diff
@@ -1,5 +1,20 @@
 import { component$ } from '@builder.io/qwik';
-import { Tooltip } from '@qwik-ui/theme-daisy';
+import { Tooltip, TooltipProps } from '@qwik-ui/theme-daisy';
+
+export const positions: TooltipProps['position'][] = [
```
gilf, line 4: Why not to use a TS enum instead?
`export enum positions { 
   TopEnd = 'top-end', 
   Top = 'top',
   ...
}`

packages/headless/src/components/tooltip/tooltip.tsx
```diff
@@ -64,15 +71,30 @@ export const Tooltip = component$(
       })
     );
 
-    useClientEffect$(({ track }) => {
+    useClientEffect$(async ({ track }) => {
+      let cleanup = () => {
+        console.log('clean up signal has not been set');
```
gilf, line 76: There are too many console.log in this file. While you probably need it for debugging, I don't see any value to commit them as part of the component implementation.

packages/website/src/routes/daisy/tooltip/index.tsx
nnelgxorz: They are already typed with the FloatingUI Placement enum.

packages/headless/src/components/tooltip/tooltip.tsx
nnelgxorz: Still a work in progress. All the console logs will be gone before we merge.

gioboa:
I think we can close it, this PR #159 solved the issue.

PR #100, opened 2023-01-28
refactor: generate the component menu from array instead of having it…

fabiobiondi:
… hard coded in the template

# What is it?

- [ x] Feature / enhancement

# Description

Generate the component menu from array instead of having it hard coded in the template


gioboa (approved):
Great 😊

PR #101, opened 2023-01-28
chore(component): remove previous Select implementation

shiroinegai:
Just removing the previous Select implementation after Gil gave his blessing. 😇


shairez:
@all-contributors please add @shiroinegai for code, ideas, bug and a11y

PR #106, opened 2023-01-29
Fix/button dynamic classes

fabiobiondi:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

The `<Button>` component now supports all Daisy modifiers (primary, secondary), sizes (xs, sm, ...) and other Daisy properties. 

A new `clsq` utility  (a TS clone of the famous `clsx` package) has been added in a new `shared` folder.
I have also added an alias in `tsconfig` in order to be easily imported everywhere:

```
import { clsq } from '@qwik-ui/shared';
```


gioboa:
It looks great 👏

PR #115, opened 2023-01-31
Feat/card: add the Card component

fabiobiondi:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
The first version of the Card Component. 

# Usage

```html
<Card>
  <CardImage src="https://daisyui.com/images/stock/photo-1606107557195-0e29a4b5b4aa.jpg" alt="shoes" />
  <CardBody>
    <CardTitle>Card Title</CardTitle>
    <p>Lorem ipsum dolor sit amet...</p>         
    <CardActions>
      <Button>Button 1</Button>
      <Button>Button 2</Button>
    </CardActions>
  </CardBody>
</Card>
```

### DIFF
--- a/packages/daisy/src/components/button/button.tsx
+++ b/packages/daisy/src/components/button/button.tsx
@@ -1,44 +1,48 @@
 import { component$, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
 import { Button as HeadlessButton } from '@qwik-ui/headless';
 import { clsq } from '@qwik-ui/shared';
+import { daisyConfig } from './daisy.config';
 
 export type HTMLButtonProps = QwikIntrinsicElements['button'];
 
 export type DaisyButtonProps = {
+  variant?: DaisyButtonVariants
+  size?: DaisyButtonSizes
+  active?: boolean;
   noAnimation?: boolean;
+  circle?: boolean;
   glass?: boolean;
   loading?: boolean;
-  circle?: boolean;
-  square?: boolean;
-  active?: boolean;
-  disabled?: boolean;
   outline?: boolean;
-  size?: 'xs' | 'sm' | 'md' | 'lg' | 'block' | 'wide';
-} & DaisyButtonModifiers;
-
-export type DaisyButtonModifiers = {
-  primary?: boolean;
-  secondary?: boolean;
-  accent?: boolean;
-  info?: boolean;
-  success?: boolean;
-  warning?: boolean;
-  error?: boolean;
-  ghost?: boolean;
-  link?: boolean;
+  square?: boolean;
 }
 
+export type DaisyButtonVariants = 'primary' | 'secondary' | 'accent' | 'info' | 'success' | 'warning' | 'error' | 'ghost' | 'link' | 'disabled';
+export type DaisyButtonSizes = 'xs' | 'sm' | 'md' | 'lg' | 'block' | 'wide';
 export type ButtonProps = HTMLButtonProps & DaisyButtonProps ;
 
 export const Button = component$(
   (props: ButtonProps) => {
     const {
-      class: cls,
-      primary, secondary, accent, info, success, warning, error, ghost, link, outline, active, disabled, glass, loading, noAnimation,
-      circle,
-      square,
       size = 'md',
-      ...rest } = props;
+      variant = 'primary',
+      circle,
+      active,
+      class: cls,
+      disabled,
+      glass,
+      loading,
+      noAnimation,
+      outline,
+      square,
+      ...rest
+    } = props;
+
+    const {
+      variants,
+      sizes,
+      options
+    } = daisyConfig
 
     return (
       <HeadlessButton
@@ -46,31 +50,17 @@ export const Button = component$(
         class={
           clsq(
             'btn',
+            variants[variant],
+            sizes[size],
             {
-              // modifiers
-              'btn-primary': primary,
-              'btn-secondary': secondary,
-              'btn-accent': accent,
-              'btn-info': info,
-              'btn-success': success,
-              'btn-warning': warning,
-              'btn-error': error,
-              'btn-ghost': ghost,
-              'btn-link': link,
-              'btn-active': active,
-              // daisy props
-              'btn-outline': outline,
-              'btn-disabled': disabled,
-              'glass': glass,
-              'loading': loading,
-              'no-animation': noAnimation,
-              'btn-circle': circle,
-              'btn-square': square,
-              // size
-              'btn-xs': size === 'xs',
-              'btn-sm': size === 'sm',
-              'btn-md': size === 'md',
-              'btn-lg': size === 'lg',
+              [options.active]: active,
+              [options.outline]: outline,
+              [options.disabled]: disabled,
+              [options.glass]: glass,
+              [options.loading]: loading,
+              [options.noAnimation]: noAnimation,
+              [options.circle]: circle,
+              [options.square]: square,
             },
             cls
           )}
--- a/packages/daisy/src/components/card/card-image.tsx
+++ b/packages/daisy/src/components/card/card-image.tsx
@@ -1,4 +1,4 @@
-import { component$, HTMLAttributes, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
+import { component$, HTMLAttributes, QwikIntrinsicElements } from '@builder.io/qwik';
 import {
   CardImage as HeadlessCardImage,
 } from '@qwik-ui/headless';
--- a/packages/daisy/src/components/card/card-image.tsx
+++ b/packages/daisy/src/components/card/card-image.tsx
@@ -1,0 +1,14 @@
+import { component$, HTMLAttributes, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
+import {
+  CardImage as HeadlessCardImage,
+} from '@qwik-ui/headless';
+export type HTMLImgProps = QwikIntrinsicElements['img'];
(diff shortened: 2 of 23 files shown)


packages/daisy/src/components/card/card.tsx
```diff
@@ -0,0 +1,18 @@
...
+import {
+  Card as HeadlessCard,
+} from '@qwik-ui/headless';
+import { clsq } from '@qwik-ui/shared';
+
+export type CardProps = HTMLAttributes<HTMLElement>;
+
+export const Card = component$(
+  (props: CardProps) => {
+    const { class: cls, ...rest } = props;
```
GregOnNet, line 11:
```suggestion
    const { class: cssClasses, ...rest } = props;
```

packages/daisy/src/components/card/card-image.tsx
```diff
@@ -0,0 +1,14 @@
+import { component$, HTMLAttributes, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
+import {
+  CardImage as HeadlessCardImage,
+} from '@qwik-ui/headless';
+export type HTMLImgProps = QwikIntrinsicElements['img'];
+
+type CardImageProps = HTMLAttributes<HTMLElement> & HTMLImgProps;
```
GregOnNet, line 7: 🔥 That's so cool. I was looking for a type-safe way to provide native HTML overrides to our components.

packages/website/src/routes/docs/headless/card/card.css
```diff
@@ -0,0 +1,23 @@
+.custom-card {
```
GregOnNet, line 1: **Question:** When do we decide to add css to headless components.

I was asking this myself for the Dialog-Component since it has a backdrop and I needed to specify width and height of the dialog.

packages/daisy/src/components/card/card-image.tsx
```diff
@@ -0,0 +1,14 @@
+import { component$, HTMLAttributes, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
```
gioboa, line 1: remove the unused import pls

packages/headless/src/components/card/card-image.tsx
```diff
@@ -0,0 +1,14 @@
+import { component$, HTMLAttributes, QwikIntrinsicElements, Slot } from '@builder.io/qwik';
```
gioboa, line 1: remove the unused import pls

packages/headless/src/components/card/card.tsx
```diff
@@ -0,0 +1,15 @@
+import { component$, HTMLAttributes, Slot, useStylesScoped$ } from '@builder.io/qwik';
```
gioboa, line 1: remove the unused import pls

packages/daisy/src/components/button/button.tsx
```diff
@@ -1,76 +1,66 @@
...
+      variants,
+      sizes,
+      options
+    } = daisyConfig
 
     return (
       <HeadlessButton
         {...rest}
         class={
           clsq(
```
GregOnNet, line 51: Maybe not related to this PR.

I read that `clsq` is a rewrite of `clsx` in TypeScript.
However, it is hard to guess, what `clsq` does.
Even reading the implementation of `clsq` does not help quickly, since it is optimized code.

Can we come up with a better name here?
Additionally, I would add a JSDoc comment, clarifying what the code  of  `clsq` does.
We can provide some Unit Tests as well.

```suggestion
          applyCssClasses(
```

I am glad to help here, if you want, since my comment is a bit out of scope for this issue.

packages/daisy/src/components/card/card-image.tsx
fabiobiondi: Ok fixed

packages/headless/src/components/card/card.tsx
fabiobiondi: ok fixed

packages/headless/src/components/card/card-image.tsx
fabiobiondi: ok fixed

packages/daisy/src/components/button/button.tsx
fabiobiondi: Yes you're right.
1. I have thought  to use `clsq` because `clsx` is "famous" lib in the React and other ecosystems.
2. I updated the code from ES5 to TS but since I have changed `arguments` with `rest` operators, var with `let` and so on there could be (even I don't think) some cases that doesn't work like the original one
3. So I also think we should integrate unit tests but @gioboa tell me that we are writing tests in this phase and we'll add them later.
4. I also think  this utility (or something similar) should be part of the Qwik core.

Anyway, in the meanwhile I can rename `cls`. `applyCssClasses` is not bad but other frameworks use:

* Angular: `[ngClass]="..."`
* Vue: `<div :class="{}"`
* Astro: `<div class:list={}`

What do you think about:
`qClass`
or 
`classList` (like the DOM property)
?

packages/daisy/src/components/card/card.tsx
fabiobiondi: Ok thank you for the advice.
I renamed it to `className` to looks like the DOM property name and use the same name already used by another component: `headless\..\tabs.tsx`

packages/website/src/routes/docs/headless/card/card.css
fabiobiondi: I have exact the same doubts.
First I decided to added multiple examples in the headless doc: with or without custom classes: I need to try if the headless component really works fine with no classes, by default) but also that can be easily customized.

Anyway there are some  headless components that don't need styles/classes to work (i.e. a Button or the Card that just to display content ).
In other (complex) use cases just like a dropdown or a dialog I think that we should embed some styles in the component itself otherwise the component doesn't work as expected by default.
In the Dialog, for example, I think we should embed the CSS it needs to display it in overlay (with some parameters like centered, fullscreen or something else to help users to easily use it in the most common scenarios ) but also give to the users the possibility to override them adding custom classes / inline styling to each element of the Dialog (backdrop, the dialog panel, title and so on)

gioboa: This headless flow makes sense 👍

shiroinegai: I have some thoughts on what "headless" means to the end user.

Too often, when I look at other UI libraries, say Bootstrap, they come with a ton of pre-configured styles/options which tend to be an additional burden for developers to overwrite or introduce extra complexity that they should be aware of. This makes it harder for users to compose their own complex components because they can and likely will get confused why their custom styles or functionality doesn't work the way they expect.

Honestly, I think the only styles that should be added to headless components are the ones that are _strictly_ for functionality/behaviour-related purposes. That also goes for options like variants that are meant to be "convenient" for the user.

For example, in the current headless Card implementation, the `variant` prop is provided only in the Daisy implementation and that's good because as a user, when I import the headless Card component, I'm expecting only semantic markup and structure that I can freely style and then incorporate into my own design system with my own `variant` prop implementation.

I think what users expect for a headless component is for it to be just like a native HTML element, but without the styling restrictions. Libraries like Headless UI, Reach UI and Radix UI have raised the bar and included accessibility out of the box and I think Qwik UI should keep to the expectations that those libraries have set.

Using an analogy, I view headless components like a headless CMS. A headless CMS just provides me the structure of the data I need to bring content to a bespoke front-end. I'd like my headless components to also simply provide me the semantic structure and accessibility behaviours I need to create my own bespoke design system of components.

fabiobiondi: Interesting discussion.  I agree @shiroinegai.

So, if we talk about the dialog component that  @GregOnNet is creating, the headless version should:

- set the dialog  in overlay? or not even that?
( it should set the display property to  fixed/absolute style? and probably its default size?)

- handle the backdrop style?
for example, should it handle a default background color? 

- but I think we are all agree that dialog title, content and footer/actions should just contains the HTML and nothing about styling.


The problem is that it's very hard to imagine the final result for end-users that read the documentation.
And probably it will be very hard to skin for people that has not enough CSS knowledge.

So I think the  key is a good documentation with nice examples:
In fact, if headless components will be totally without styles we should always show one or more examples with custom classes showing the final result, just for inspiration or to copy and quickly update the CSS

shiroinegai: Looking at the headless Dialog implementation from @GregOnNet , since it returns a native `dialog` element, I think it's best to not include any additional styling. You can see the default browser styles for a `dialog` on MDN and it has a 3px black border with 16px padding.

I think the point with headless components isn't to look nice out of the box, but to behave correctly even though they may look absolutely ugly. Radix UI does a really fantastic job using a `styles.css` tab in their docs. If you import just the component parts, it really comes unstyled. You can see how their Dropdown component looks with just the markup alone as demo-ed by one of their maintainers here: https://youtu.be/pcMYcjtWwVI?t=535

> The problem is that it's very hard to imagine the final result for end-users that read the documentation.
And probably it will be very hard to skin for people that has not enough CSS knowledge.

I think this is why we'll be having the Daisy and Material implementations. Those who aren't keen on deeply customising the look of their components can just import those implementations instead. 

These implementations should be composed with our headless components, just that we are also porting over the design (and perhaps even props) that users of those libraries are familiar with. And especially for Daisy, it will also be an enhanced experience for users because it is now composed of Qwik UI's accessible components.

Familiar look, but now accessible!

PR #117, opened 2023-01-31
style: renamed cls in className

fabiobiondi:
# What is it?

- [x] Feature / enhancement
- [] Bug
- [ ] Docs / tests

# Description

Just renamed the `cls` property in `className`

### DIFF
--- a/packages/daisy/src/components/button-group/button-group.tsx
+++ b/packages/daisy/src/components/button-group/button-group.tsx
@@ -6,9 +6,9 @@ export type ButtonGroupProps = HTMLAttributes<HTMLElement>;
 
 export const ButtonGroup = component$(
   (props: ButtonGroupProps) => {
-    const { class: className, ...rest } = props;
+    const { class: classNames, ...rest } = props;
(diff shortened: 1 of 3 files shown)


shairez:
thanks @fabiobiondi !

the only issue I see with this is the confusion with React's `className` and the deprecation of it in Qwik

what do you think?

GregOnNet:
Maybe it is simply `classNames` since the developer can pass a bunch of classes.

fabiobiondi:
I had the same doubts 😅 
We should decide a "name" and use it everywhere.

I used `className` because it reminds the DOM property and its just like using `class`.
But as @shairez  said it may create confusion (even if it's just for internal use and invisible to end-users).

`classNames` seems ok ( custom, plurals and should be clear what it  contains)

So should I change it everywhere? 😅

---

To give some context ,  the problem is  generated when the  destructuring is used  with the `class`  key:

```
const { class } = props ; <-- ❌
```


```
const { class: something } = props ✅
```

Issue #108, opened 2023-01-29
Prefer "class" over "className"


GregOnNet:
## Reproduction

- Starting the website

```
cd packages/website
npm start
```

- The following warning is shown

<img width="522" alt="CleanShot 2023-01-29 at 21 23 20@2x" src="https://user-images.githubusercontent.com/444278/215353933-7e927187-40a4-4088-9b88-bacb2174d02c.png">

## Proposed solution

We might want to replace `className` with `class`.

## Final thoughts

- [x] I am willing preparing a PR for it.

shairez:
thanks @GregOnNet !
go for it! 💪🙏

GregOnNet:
I encounter an error executing `npx nx storybook core` (stated in contribution guideline).
I am using Node 18.

![image](https://user-images.githubusercontent.com/444278/215984836-6fbe885c-b819-4f3b-9bba-45c87d4445ab.png)

So I executed `npx sb@next automigrate` like stated in the error.
This ran the following migrations: 
- 'newFrameworks', 
- 'mainjsFramework', 
- 'addReact'
- 'eslintPlugin'

After the migration storybook does not start.

![image](https://user-images.githubusercontent.com/444278/215985402-6b369756-ca1d-47bd-9323-f023d1e6a2a2.png)

GregOnNet:
So I verified that my change does not break anything by:

```
cd packages/website
npm start # clicking through the components 🤗
```

Balastrong:
I see there are no more `className` props in the project, maybe we can close this issue? cc @gioboa 

Actually, there are some variables here and there called `classNames`, just in case we want to rename those as well, but they shouldn't be a problem so far.

gioboa:
Yes, thanks.

PR #120, opened 2023-02-01
fix(jsx): replace className with class

GregOnNet:
# What is it?

- [x] Enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Description can be found at #108 

# Use cases and why

A `jsx`-deprecation warning was shown

# Checklist:

- [-] I have made corresponding changes to the documentation
- [-] Added new tests to cover the fix / functionality

`-` means not applicable

### DIFF
--- a/packages/daisy/src/components/tabs/tab.tsx
+++ b/packages/daisy/src/components/tabs/tab.tsx
@@ -16,7 +16,7 @@ export const Tab = component$(
         class={`tab ${isBordered ? 'tab-bordered' : ''} ${
           isLifted ? 'tab-lifted' : ''
         }`}
-        selectedclass="tab-active"
+        selectedClassName="tab-active"
         {...props}
       >
         <Slot />
--- a/packages/daisy/src/components/tabs/tab.tsx
+++ b/packages/daisy/src/components/tabs/tab.tsx
@@ -16,7 +16,7 @@ export const Tab = component$(
         class={`tab ${isBordered ? 'tab-bordered' : ''} ${
           isLifted ? 'tab-lifted' : ''
         }`}
-        selectedClassName="tab-active"
+        selectedclass="tab-active"
         {...props}
       >
         <Slot />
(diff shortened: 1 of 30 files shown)


GregOnNet (commented):
reviewed

shairez:
so "core" is going to be depracated in favor of `theme-daisy`
sorry if it was not communicated earlier

So I think we need to maybe change its name to `deprecated-core` or something

GregOnNet:
Ah, thanks.

Since most of it is in core. We could also drop this PullRequest. What do you think?

If you agree, feel free to close this PR

shairez:
I think we can merge it because it has the fixes you did for the other packages 

If you'd like to rename the folder as part of your change I'll wait for that before merging

packages/daisy/src/components/tabs/tab.tsx
```diff
@@ -16,7 +16,7 @@ export const Tab = component$(
         class={`tab ${isBordered ? 'tab-bordered' : ''} ${
           isLifted ? 'tab-lifted' : ''
         }`}
-        selectedClassName="tab-active"
+        selectedclass="tab-active"
```
shairez, line 19: I think that in this case, it makes sense to call it `selectedClassName` (just to make sure the prop name is clear enough that it's not a ES Class or something)

WDYT?

GregOnNet: I would go for `classActive`.

```tsx
<Tab class={/* ... */} classActive="tab-active"/>
```

To me, this looks convenient.

GregOnNet: Ha, just noticed, that this came from my search and replace.

I should revert this, and we can have a discussion on API-Surface for components later.

GregOnNet:
@shairez renamed slectedClass to `selectedClassName`. 👍

shairez:
thanks @GregOnNet !

I see that @gioboa already merged his PR renaming the core package so this introduced conflicts

I don't want you spending more time on this as it's not critical (we'll be moving stuff out of the core package soon anyway)

So I'll close this PR for now, again - sorry for not communicating about the `core` earlier and thanks for investing your time in trying to make this project better! 🙏

gioboa:
@GregOnNet Sorry for the miscommunication 😔 thanks for your contribution 🙏

PR #140, opened 2023-02-02
feat(component): add draft Popover component

fabiobiondi:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
This is just a draft of the Popover component to discuss about the used approach and check if there are better ways to handle the communication between its children components


fabiobiondi:
Hi, I would like to explain some technical details and motivations about some technical choices and strategies I have used to build the `Popover` component:

Here you can see what is our goal and how the end user should use the component:

1) Example with HTML elements used  inside the Trigger and the Content children::

```html
<Popover>
  <PopoverContent>My Content here</PopoverContent>
  <PopoverTrigger> ➡️ CLICK ME ⬅️</PopoverTrigger>
</Popover>
```

2) Example that use custom components as trigger (`Button`) and content (`Calendar`):

```html
  <Popover>
        <PopoverContent><Calendar /></PopoverContent>
        <PopoverTrigger >
          <Button> CLICK </Button>
        </PopoverTrigger>
   </Popover>
```

It's very important to note that the `Popover` component handles one `Slot` only.
But I need to get both children references  inside it in order to apply custom styles, event listeners and do some stuff (for example, the Trigger component, when  clicked, should display/render the Content component so they must communicate).

The problem is that Qwik does not provide API to do this out of the box. 
I mean it seems there are no ways to get a specific reference to every element passed as Slot.

Here you can see the current source code of the main `Popover` component.
Read comments to better understand what is does:

```tsx
/*
the context is used to allow children (`PopoverContent` and `PopoverTrigger` to communicate with the parent,
  since these components are not directly visible to the Popover that just create them by using  `Slot` 
*/
interface ContextProps { isOpen: boolean }
export const myContext = createContext<ContextProps>('my-context');

export default component$(() => {
  // context is used to save the state of the Popover, opened or not
  const state = useStore({ isOpen: false });
  useContextProvider(myContext, state);
  
  // the wrapper ref is just used  to limit the selection with querySelector. 
  // So I can use multiple instances of this component in the same page
  const wrapper = useSignal<HTMLElement>();

  useClientEffect$(() => {
     /**
      Since this component has no direct access to its children (trigger and content) I need a way to refer them.
     I did not find other ways except the use of `querySelector`:  
      Each child component have a role inside its template (you can find inside their source code below)
       and I can get their HTML reference by using that.
     NOTE: probably I should use a different role `name` or a different way to get them (by className?). 
     Any suggestions is welcome :)
     **/
      const content = wrapper.value?.querySelector<HTMLElement>('[role="content"]');
      const trigger = wrapper.value?.querySelector<HTMLElement>('[role="trigger"]');
      
      /*
       Now I just Use floating-ui to set the right "content" position 
       and this is the place where I need to use the  references to children (trigger and content) 
    */
      return autoUpdate(trigger!, content!, () => {
        computePosition(
          trigger!,
          content!,
          { middleware: [flip(), shift()] }
        ).then(({x, y}) => {
          Object.assign(content!.style, {
            left: `${x}px`,
            top: `${y}px`,
          });
        });
      });
  });

  // This component just render its children
  return (
    <div ref={wrapper}>
      <Slot />
    </div>
  )
});

```


---

## Child: `PopoverTrigger`:


The Trigger component is used to define the UI and the Popover behaviors (if opening it by clicking, at rollover, ...).
It can contains valid JSX value: DOM elements or Components:

```html

      <Popover>
        <PopoverContent>....</PopoverContent>
        <PopoverTrigger> ➡️ CLICK ME ⬅️ </PopoverTrigger>
      </Popover>

```

The `PopoverTrigger` component just uses `<Slot>` to display its content but I wrap it with a `span` because:

1) I apply a `role` attribute, used by the parent (`Popover`) to get its reference (as explained in the comment above)
2) it should display the `PopoverContent` component when clicked (in the future it will support the hover too, to act as a Tooltip)  setting  the  `isOpen`  state (defined in the parent, `Popover`)  to `true`

IMPORTANT: Since `PopoverTrigger` is not a direct child of the `Popover` ( but used as `Slot`) I cannot use callbacks / props to allow the communication and that's the reason I use a context to share the `isOpen` prop 

```tsx
export const PopoverTrigger = component$(
  () => {
    const ctx = useContext(myContext);

    return (
      <span
        role="trigger"
        class="popover-trigger"
        onClick$={() => {
          ctx.isOpen = !ctx.isOpen
        }}
      >
        <Slot />
      </span>
    );
  }
);


```

Since this component triggers the event to open the content (now the click event but it should handle mouseOver too) it should  also be used to define it . We may add a prop:

```tsx

// popover component
interface PopoverTriggerProps {
  triggerEvent?: 'click' | 'mouseOver' | '...';
}

export const PopoverTrigger = component$(
  (props: PopoverTriggerProps) => {
  
  ... missing code here...
```

and it may be used in this way:

```tsx
    <Popover>
        <PopoverContent>...</PopoverContent>
        <PopoverTrigger triggerEvent="mousover"> ➡️HOVER ME ⬅️</PopoverTrigger>
      </Popover>
      ```

This should work even if  I would prefer apply this property to the parent `Popover` itself and pass down to the `PopoverTrigger` as shown below:

```tsx
<Popover triggerEvent="mouseover">
```

but how the `PopoverTrigger` can know what property has been sent to the `Popover` since it's not a direct child ?
Maybe using `Context` again? 
The `Popover` may store this value in a local state and use  the context to pass down the value to the children .
Still have to think about it.

---

## Child: `PopoverContent`:

This component just display the `Popover` content using `Slot`.
Furthermore it applies dynamic CSS classes to display or hide it (getting the `isOpen` state from context)

```tsx
export const PopoverContent = component$(
  () => {
    useStylesScoped$(styles);
    const ctx = useContext(myContext);

    return (
      <div
        role="content"
        class={clsq(
          'popover-content',
          {
            open: ctx.isOpen,
            close: !ctx.isOpen,
          }
        )}
      >
        <Slot />
      </div>
    );
  }
);


```

What do you think about this approach? 
I have not found other cool and cleanest ways to achieve the same result.

I have also created a [StackBlitz](https://stackblitz.com/edit/qwik-starter-mqpgem?file=src/components/popover/popover.tsx)  if you want to play with it

shiroinegai:
I'm not available to look through the implementation yet, but there's one particular issue I noticed.

The `role` property should not be used in this manner. It is specifically meant for ARIA roles only.

See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

fabiobiondi:
HI @shiroinegai, 
yes  you're right and I imagined that.
Thank you for the link. I read it now.

Anyway this is just a draft and before spending time on the details 
I would like to be sure there are no better ways to achieve this goal.

packages/headless/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface ContextProps { isOpen: boolean }
+export const myContext = createContext<ContextProps>('my-context');
+
+// popover component
+interface PopoverProps {
+  alignments?: 'auto' | 'left' | 'top' | '...';
+}
+
+export const Popover = component$((props: PopoverProps) => {
+  const state = useStore({ isOpen: false });
```
gilf, line 21: Please replace with useSignal

packages/headless/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  useContextProvider(myContext, state);
+
+  const wrapper = useSignal<HTMLElement>();
+
+  useClientEffect$(() => {
```
gilf, line 27: We should avoid client effect. There are better  alternatives for it

packages/headless/src/components/popover/popover.tsx
fabiobiondi: are you sure @gilf ? 
It seems I cannot update the `isOpen` context property from a children component anymore (it lose the reactivity).
Or I wrong? In fact I see in the documentation it uses `useStore` to read and writea simple `counter`

packages/headless/src/components/popover/popover.tsx
fabiobiondi: I have tried the following code but it's invoked about 14 times for each Popover instance 😅

```tsx
  useOnDocument('load', $((event)=> {
    console.log('onload',event)
  }))
```

I don't understand if it's a bug (or a feature 😅)  but I cannot use it to initialize the popover with floating-ui.

Any ideas?

fabiobiondi:
I have done a huge refactoring to the popover component and add some new features too:
* autoplacement
* animations
* clickOutside
* custom alignments
*  offset
* choose between click and hover triggers 
* and i'm doing some experiments to create a controlled component from the parent (the parent can decide when the popover should be closed or opened)

packages/headless/src/components/popover/popover.tsx
shairez: @gilf I think we need `useClientEffect$` because the query needs to happen *after* the first render occurred

fabiobiondi:
I have done another refactoring to remove all querySelectors and use references only.

gioragutt:
Question about the hierarchy of popover and it's children.
To my understanding, the `Popover` must have `PopoverContent` and `PopoverTrigger` as it's two children.

Could this be instead in a more directive-style approach, sort of like how `@angular/material`'s [`<mat-menu>`](https://material.angular.io/components/menu/overview) or `MUI`'s [`Menu`](https://mui.com/material-ui/react-menu/#positioned-menu) is implemented?

Where the trigger is acquired using something like a `ref`, and the `PopoverContent` would contain the logic? In practice this would remove two of the components, since `PopoverContent` would contain all the logic, and would in practice turn into `Popover` itself, and the `PopoverTrigger` would not be a component but instead would be a simple ref.

PR #198, opened 2023-02-23
fix: update config to enable cypress component testing

dmitry-stepanenko:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
Cypress component testing wasn't working for libs. To fix this, should use `mode: 'lib'` when running cypress tests. Additionally, it is required to configure `server.fs.allow` in the `vite.config.ts`

### DIFF
--- a/packages/headless/src/components/popover/popover.tsx
+++ b/packages/headless/src/components/popover/popover.tsx
@@ -52,7 +52,10 @@ interface PopoverProps {
 }
 
 export const Popover = component$((props: PopoverProps) => {
-  const { triggerEvent = 'click', onUpdate$, disableClickOutSide } = props;
+  const triggerEvent = props.triggerEvent ?? 'click';
+  const onUpdate$ = props.onUpdate$;
+  const disableClickOutSide = props.disableClickOutSide;
+
   const wrapperRef = useSignal<HTMLElement>();
   const triggerRef = useSignal<HTMLElement>();
   const contentRef = useSignal<HTMLElement>();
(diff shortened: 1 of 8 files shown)


packages/headless/src/components/popover/popover.tsx
```diff
@@ -52,7 +52,10 @@ interface PopoverProps {
 }
 
 export const Popover = component$((props: PopoverProps) => {
-  const { triggerEvent = 'click', onUpdate$, disableClickOutSide } = props;
+  const triggerEvent = props.triggerEvent ?? 'click';
+  const onUpdate$ = props.onUpdate$;
+  const disableClickOutSide = props.disableClickOutSide;
```
dmitry-stepanenko, line 57: original expression stopped working after migration to latest qwik

shairez: what do you mean by "original expression"?

dmitry-stepanenko:
```js
const { triggerEvent = 'click', onUpdate$, disableClickOutSide } = props;
```

there's a new error `props` has already been defined. has something to do with optimizer, I believe, will try to replicate and create a ticket in qwik later, if it's reproducible

shairez: OK, nice catch!
yeah an issue is a great idea
thanks Dima!

shairez:
great job @dmitry-stepanenko ! 🙏👏

PR #227, opened 2023-03-05
feat: add PR linter bot

gioboa:
# What is it?

- [X] Feature / enhancement

# Description

In order to run few checks against PRs, I created this bot


shairez:
@gioboa thanks!

I'd do the following changes - 

rename lint-pr to "security-check" because it's mainly about security and the `danger` package right? 

Plus, why not adding it under the existing `test` action? we have a lint step there (that's why I offered to rename, not to collide), and this should also run as part of the tests no?

gioboa:
Yes, you're right. Let's move it in the same action. Would you like to do this change pls?

Issue #239, opened 2023-03-13
[Toast] A11y: Use `output` instead of div


GrandSchtroumpf:
Using the `output` html element for a toast instead of a div improve a11y as it will be read by the screen reader when it's displayed on the page without the need to get focus. By default the `output` element has the role `status`, but for more time-sensitive information the role `alert` can be set.
- [output element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output)
- [role status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role)
- [role alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)

PR #240, opened 2023-03-13
feat(toast) Use html output element

GrandSchtroumpf:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
As described in #239, using `output` html element improve the a11y to announce a message after a user action or time-sensitive information (which is usually what a Toast is for).
This is a first PR to start playing with the project a little.

# Checklist:

- [ ] Added new tests to cover the fix / functionality (Not applicable)

### DIFF
--- a/packages/primitives/src/lib/toast/toast.tsx
+++ b/packages/primitives/src/lib/toast/toast.tsx
@@ -4,12 +4,18 @@ export type ToastProps = {
    */
   label?: string;
   class?: string;
+  /**
+   * Aria role: default 'status'.
+   * @description The toast uses the html element output, which has implicit role 'status'
+   * For information that requires an immediate attention of the user use 'alert'.
+   */
+  role?: 'alert' | 'status';
 }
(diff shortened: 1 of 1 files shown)


gilf:
Thanks for the change.
We, the team, need to discuss this before approval because this is one way to implement a11y on a toast.
Another way can be found here: https://react-spectrum.adobe.com/react-aria/useToast.html#anatomy.

GrandSchtroumpf:
Sure no problem. Though I would recommend not to set the role to alert by default. Not all messages in a toast are time sensitive (ex: notifications, confirmation from the server, ...). That's why I think the status role (provided by the output element by default) should be used for most cases and let the user switch to alert for time sensitive message (errors, invalid form, ...)

thejackshelton:
Closing due to inactivity, but there's a good chance this will be used in the future Toast component of Qwik UI headless!

Issue #254, opened 2023-03-25
docs: fix missing key in a for loop menu


tleperou:
# Summary

Fix

```sh
JSXError: Multiple JSX sibling nodes with the same key.
This is likely caused by missing a custom key in a for loop
```

PR #262, opened 2023-03-30
#254 Docs menu key fix

PatrykGodlewski:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Screenshots/Demo

# Checklist:

### DIFF
--- a/apps/website/src/components/menu/menu.tsx
+++ b/apps/website/src/components/menu/menu.tsx
@@ -1,7 +1,15 @@
-import { $, component$, PropFunction, useContext } from '@builder.io/qwik';
+import {
+  $,
+  component$,
+  PropFunction,
+  useContext,
+  useStore,
+  useVisibleTask$,
+} from '@builder.io/qwik';
 import { Link } from '@builder.io/qwik-city';
 
 import { APP_STATE } from '../../constants';
+import { AppPath } from '../../types';
 import { CloseIcon } from '../icons/CloseIcon';
 
 type Props = {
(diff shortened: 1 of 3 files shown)


PatrykGodlewski:

apps/website/src/components/menu/menu.tsx
```diff
@@ -10,65 +18,77 @@ type Props = {
...
-    {
-      label: 'Checkbox',
-      path: `/docs/${appState.theme.toLowerCase()}/checkbox`,
-    },
-    {
-      label: 'Progress',
-      path: `/docs/${appState.theme.toLowerCase()}/progress`,
-    },
-  ];
+  const menu = useStore<{ value: AppPath[] }>({
```
gioboa, line 21: I don't understand why you put the menu in a state. Is there any reason?

PatrykGodlewski: In that solution, I utilized state to update the menu directly on the client after theme was set. 

But your question made me think and I revisited my approach and made some improvements in the next commit. I noticed that the server-side state for the theme was always "NOT_DEFINED," so I decided to update the theme state on the server using useLocation instead of waiting for a client-side state update. Additionally, I prevented  flickering menu which was caused by the state on every route change being defaulted to "NOT_DEFINED" 

With these changes, I believe that the new solution I have implemented should effectively address the issue at hand. Please let me know if you have any further suggestions or feedback.

gioboa:
@all-contributors please add @PatrykGodlewski for bug, code

gioboa:
Thanks for your help 👍

PR #322, opened 2023-05-21
fix(select): content blocking aria-expanded and "flickering" on lazy execution

thejackshelton:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Issues fixed:
- The content was blocking our isExpanded signal from toggling inside our SelectTrigger component.
- listbox would not be in the correct computed position when the floating-UI library is lazily executed on interaction.

# Use cases and why

Writing this down in case any docs changes need to be made or there are any potential drawbacks to this implementation.

## isExpanded
Believe I figured out the issue where the content was blocking our isExpanded signal from toggling inside our SelectTrigger component.

It has to deal with event bubbling. From what I understand clicking on the onClick$ handler will now prevent any parent events from firing as well. Normally it would "bubble" all the way up to the HTML element.

https://javascript.info/bubbling-and-capturing

Inside of **SelectTrigger** I added

```tsx
onClick$={(e) => {
  e.stopPropagation();
  contextService.isExpanded.value = !contextService.isExpanded.value;
}}
```

## Flickering between select dropdown

Another thing, inside of **SelectRoot**

Dealing with the issue where momentarily the **listbox would not be in the correct computed position** when the floating-UI library is lazily executed on interaction.

I believe the issue stems from the computePosition function from floating UI. It's async and returns a promise as shown in the documentation here

https://floating-ui.com/docs/computeposition 

I added the following code and that seems to have "fixed" the flickering for now.

When all three of these conditions are true it will be hidden at first. I tried setting it to display instead of visibility but there seemed to be some weird issue with the padding not registering. I realize that this probably has to deal with the isExpanded context styles being applied first, which is why it seems to be fine when using visibility instead.

The then method is how I make the listBox visible again after the promise has resolved.

```tsx
const trigger = track(() => contextService.triggerRef.value);
const listBox = track(() => contextService.listBoxRef.value);
const expanded = track(() => isExpanded.value);

if (expanded && trigger && listBox) {
  // maybe setting display, position, and z-index here instead of inside our ul tag? I'm not sure.
  listBox.style.visibility = 'hidden';

 // update: I switched it to await in the commit instead of then.
  updatePosition(trigger, listBox).then(() => {
    listBox.style.visibility = 'visible';
  });
}
```

**First Fix:**

[Here is before](https://gyazo.com/ec5360ea1f11e50422016bc3f40b0859)

[Here is after](https://streamable.com/krctr8)

**Second Fix:**
[Here is before](https://www.awesomescreenshot.com/video/17589308?key=60315f2af6d4cae0904c9c6e696f1d7e)

[Here is after](https://discord.com/channels/990511757091033108/1108949840689369098/1109328202477879437)

# Checklist:


packages/kit-headless/src/components/select/select.tsx
```diff
@@ -71,7 +72,11 @@ export const SelectRoot = component$(
       const expanded = track(() => isExpanded.value);
 
       if (expanded && trigger && listBox) {
```
GregOnNet, line 74: Maybe split this up in 2 if-statements.

```suggestion
     if(!trigger || !listBox) return;

      if (expanded === true) {
```

GregOnNet: There is also an if-Statement below, that could be merged here after the "if-statement-split" is done.

```ts
if (expanded === true) {
  listBox.style.visibility = 'hidden';

  await updatePosition(trigger, listBox);

  listBox.style.visibility = 'visible';

  listBox.focus(); // from the if-statement below
}
```

packages/kit-headless/src/components/select/select.tsx
```diff
@@ -113,9 +118,9 @@ export const SelectRoot = component$(
   }
 );
 
-export interface TriggerProps extends StyleProps {
+export type TriggerProps = {
   disabled?: boolean;
```
GregOnNet, line 122: Do we still need this?
I guess `disabled`is part of `QwikIntrinsicElements['button']`.

packages/kit-headless/src/components/select/select.tsx
```diff
@@ -128,7 +133,8 @@ export const SelectTrigger = component$(
         ref={ref}
         aria-expanded={contextService.isExpanded.value}
         disabled={disabled}
-        onClick$={() => {
+        onClick$={(e) => {
+          e.stopPropagation();
```
GregOnNet, line 137: Question: Why do we need to use `stopPropagation`?

I believe there is a good reason for this.
Maybe we can extract this inline-block to a function and add a comment providing a bit more context why this is needed.

GregOnNet (changes requested):

packages/kit-headless/src/components/select/select.tsx
thejackshelton: I believe so. I added this with marcus on a short call.

packages/kit-headless/src/components/select/select.tsx
thejackshelton: So when there is any content that's currently in the select and you click on that content the select component will not toggle isExpanded.

I believe this has something to do with a parent event "intercepting" our other event. Because when I added e.stopPropagation it would actually toggle isExpanded when clicking on the content.

The only other solution I could think of for this would be to add pointer-events: none in CSS, but I think that would cause a lot of confusion on the consumer side as well as potential issues.

packages/kit-headless/src/components/select/select.tsx
thejackshelton: This seems a lot more declarative/explicit. Good catch. Let me make sure the functionality is the same and I'll commit the changes.

thejackshelton: Yep looks good. Gonna add the new commit

GregOnNet (approved):

thejackshelton:
Ah great. I don't know what I'm doing with git guys. If anyone knows how to remove the latest commit I made remotely in this PR let me know. Hopefully a learning experience xd

thejackshelton:
ah I think that might've got it back to what it was lol.

PR #498, opened 2023-11-08
refactor(highlight): vitePluginRequireResolve

PatrickJS:
try to run wasm on server by fixing require.resolve with a plugin


PatrickJS:
doesn't work

PR #501, opened 2023-11-11
refactor(package.json): refactor scripts for convenience pnpm workflow

maiieul:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Tried to mimic Qwik package convention with the "." notation

# Use cases and why

Using nx scripts commands doesn't feel natural at all to me as I'm too used to writing "pnpm dev", etc. I think this should be true for many other devs as well, shouldn't it?

I  tried to make the scripts as convenient as possible (e.g. `pnpm dev` instead of `pnpm website.dev` for the docs site as this is the command that will be run the most often).

I made sure every script that used the ":" notation now uses the "." notation. This is because the Qwik repository uses the "." notation and contributors to qwik-ui are likely to be contributors to Qwik as well.  

This might feel disturbing to contributors used to the current pnpm commands (e.g. it's now `pnpm dev` instead of `pnpm website` and `pnpm test.headless` instead of `pnpm test:headless`). But I believe this to be a more standard workflow and I think adapting to this change now should be worth it going forward. Also that shouldn't bother those that are used to writing the nx commands.

Tested this with the  new `pnpm preview` and `pnpm preview.cloudflare` and everything seems to work properly.

# Checklist:


thejackshelton:
Looks good to me 👍 . Because this is removing some deps and touching commands let's get a 2nd review.

maiieul:
Yeah that ci error again :sweat_smile: 

Weird I didn't have it with the first few commits.

Btw, I didn't remove any deps, just moved them.

gioboa (approved):
It looks awesome to me. thank @maiieul  you are great

PR #503, opened 2023-11-11
feat(select-listbox.tsx): added spec for printable chars

TheMcnafaha:
This commit adds both the code and tests (cypress) around adding the "Printable Characters" behavior especified by the W3C.

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

* Added logic to handle keys being pressed, "matching" any items in the select list
* Added test to ensure correct behavior

# Use cases and why

-1. Keep qwik-ui closed to the W3C specs

# Checklist:


TheMcnafaha:
Weirdly, the new test I added fail on "cold start". If you repeat the same test right afterwards, they all pass. Looking into it.

thejackshelton:
Hey MCN! I looked over the PR and see that the tests do seem to randomly be failing. Perhaps it might be a race condition issue.

I see that when we hover over options we're actually focusing them, I think what we want to do here instead, is have a "focus indicator", the combobox is a good example of that. (you can see that focus is remains on the input)

If you see the Radix Primitives select they are doing the same thing. 

Would you like to show off what you've done in discord tomorrow?

gioboa:
@TheMcnafaha are you still working on this PR or is it abandoned?

TheMcnafaha:
> @TheMcnafaha are you still working on this PR or is it abandoned?

I just need to update the docs and some minor refactors. Should be done in two days.

TheMcnafaha:
I think I will just make a new pr lol

Issue #519, opened 2023-11-18
Bug: Cannot resolve symbol s_c7PiYJ1s0GU... Error code (31) - Namespace conflicts


maiieul:
Error message:

> Cannot resolve symbol s_c7PiYJ1s0GU in {
  '02wMImzEAbk': [ 's_02wMImzEAbk', 'entry_QwikCityProvider-a7f704b4.js' ],
  ClIa90D1LOw: [ 's_ClIa90D1LOw', 'entry_AccordionTrigger-0b3712d8.js' ],
  moIlZJ0XpGY: [ 's_moIlZJ0XpGY', 'entry_AccordionTrigger-0b3712d8.js' ],
  yqEZj1cracY: [ 's_yqEZj1cracY', 'entry_AccordionContent-4d7f7f97.js' ],
  FVuPkVHsyn8: [ 's_FVuPkVHsyn8', 'entry_Slider-714836a4.js' ],
  M0tH5WLqYIQ: [ 's_M0tH5WLqYIQ', 'entry_AccordionContent-4d7f7f97.js' ],
  ZlkQX0dQ4iE: [ 's_ZlkQX0dQ4iE', 'entry_AccordionTrigger-0b3712d8.js' ],
  py4L6HieTpw: [ 's_py4L6HieTpw', 'entry_AccordionTrigger-0b3712d8.js' ],
  '0vphQYqOdZI': [ 's_0vphQYqOdZI', 'entry_RouterHead-adf3888c.js' ],
  '8gdLBszqbaM': [ 's_8gdLBszqbaM', 'entry_Link-308277b9.js' ],
  FpJ0F5de3jc: [ 's_FpJ0F5de3jc', 'entry_AccordionContent-4d7f7f97.js' ],
  HLYQug4XlU4: [ 's_HLYQug4XlU4', 'entry_AccordionTrigger-0b3712d8.js' ],
  Nk9PlpjQm9Y: [ 's_Nk9PlpjQm9Y', 'entry_GetForm-bc74dc7f.js' ],
  TxCFOy819ag: [ 's_TxCFOy819ag', 'entry_QwikCityProvider-a7f704b4.js' ],
  VKFlAWJuVm8: [ 's_VKFlAWJuVm8', 'entry_layout-f303f0c7.js' ],
  WmYC5H00wtI: [ 's_WmYC5H00wtI', 'entry_QwikCityMockProvider-1dd1a1fb.js' ],
[...]

PR #511, opened 2023-11-15
build(headless): fix large bundle size

shairez:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

fixing the large bundle size problem of the headless package

fixes #472

PR #521, opened 2023-11-20
temporary-fix(big bundle & cannot-resolve-symbol): revert to non-optimized bundling until it's fixed in qwik

maiieul:
**TLDR**:  I think it would be best to keep the unoptimized bundle until a real fix can be made in the optimizer or until we're ready for v1 if the optimizer issue hasn't been fixed by then.

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

I removed

```tsx
      output: {
        preserveModules: true,
        preserveModulesRoot: 'packages/kit-headless/src',
      },
```

from the headless-kit's vite config, because it creates the namespace issues described in #519, for which there is no great workaround in terms of DX.

This brings back the big bundle issue, but the "big bundle" isn't so big on first load anyway, and subsequent requests will be loaded from the browser cache.

To make the bundle smaller, I removed country-list-json and libphonenumber-js dependencies and all related code (which didn't receive any attention for the past 6 months). We can always add it back later or create another package for those kinds of components if the bug hasn't been fixed by then.

Performance should be okay now (11kb on first load :point_down:)

image

:warning:  I reused some of the code from #511, updated to latest qwik version, and made sure that everything works in dev and preview (the workspace dependency was causing issues to the build step). I suggest to first merge this PR and then @shairez you can work on top of it for the utils package.

# Checklist:


maiieul:
In case anyone's looking into this, I had a programming session with Shai and we think we know what's causing the namespace issues when using `preserveModules`, so we're holding off for now.

shairez:
closing this in favor of the recent fixes

PR #523, opened 2023-11-24
refactor((components) folders): move component examples from `/(components)` into parent folders (/fluffy & /headless)

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

This intermediary PR is to simplify the review process for my incoming PR that does the same change. This way you will be able to see the actual diffs for each file instead of seeing completely new files.

Although I believe everything will work fine in my following PR, this intermediary PR can easily be reverted. Please, merge this PR before reviewing the following PR, and you can revert the intermediary PR if there's any blocking problem.

### Explanation:

TLDR; when using `import.meta.glob`, it's best to **NOT** use grouped layouts as it makes working with the `useLocation()` hook (`location.url.pathname`) harder.

[Grouped layouts](https://qwik.builder.io/docs/advanced/routing/#grouped-layouts) are particularly useful for organizing the codebase logically.

But my following PR will add the `import.meta.glob` pattern with a wildcard search for retrieving the files.

Ex:

```tsx
const components: any = import.meta.glob('/src/routes/docs/**/**/examples/*.tsx', {
  import: 'default',
  eager: isDev ? false : true,
});
```

Will allow me to retrieve 

`/src/routes/docs/components/headless/modal/examples/hero.tsx` when the URL is at `/docs/headless/modal`
Or `/src/routes/docs/components/fluffy/modal/examples/hero.tsx` when the URL is at `/docs/fluffy/modal`.

In order to do so, I need to use the `useLocation()` hook in the component like that:

```tsx
  const location = useLocation();
  const componentPath = `/src/routes${location.url.pathname}examples/hero.tsx`;
```

This code can be used when there is no grouped layout (here the `/(components)` folder I removed). This is simpler and more resilient to code changes.

If we use grouped layouts, then we can't directly use `location.url.pathname`. We have to change the URL string to add `"/(components)"` in it, and we have to change the replace function every time we move files around.

# Checklist:


maiieul:
Closing, as I'm afraid this is actually a bad idea. Could potentially create unnecessary conflicts to resolve.

maiieul:
Sorry, can be closed now (needed to copy/paste the contents of my comment).

PR #525, opened 2023-11-26
chore: upgrade qwik to latest

dmitry-stepanenko:
Upgraded qwik to latest. 

There's a bug with rendering the app in cypress due to a modified path to the `@vite` asset. It was working by accident before, but stopped after this PR https://github.com/BuilderIO/qwik/pull/5234. Until this is properly fixed on the qwik side, we need to patch the `@builder.io/qwik` package so that `__cypress/src/@vite/` asset path is also recognized (simply modified the regex to match the whole `pathname` instead of its beginning)

Additionally removed the pnpm workspace configuration so that all dependencies are managed from the root package.json. Right now there's a mess with versions of dependencies, `daisyui` even has different major versions specified.


shairez:
Thanks @dmitry-stepanenko ! 🙏

PR #537, opened 2023-12-05
refactor(modal/examples): replaces fieldsets in example code

TheMcnafaha:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Better semantic HTML.

# Use cases and why

This commit makes the code closer to the W3C's recommendation, which are quoted below, in two important ways:

- 1. Removes `fieldsets` that had no legends
> The first element inside the fieldset must be a legend element, which provides a label or description for the group

[source](https://www.w3.org/WAI/WCAG21/Techniques/html/H71)

- 2. Applies the recommended pattern for *self-explanatory* forms like the ones used in the examples by replacing the `fieldsets` with `divs`, relying solely on the labels for meaning
> when [...] the individual label associated with each particular control provides a sufficient description [...] the use of the fieldset and legend elements is not required

[source](https://www.w3.org/WAI/WCAG21/Techniques/html/H71)

Alternately, we could add  `legend` tags to each `fieldset` instead of removing them, but the W3C seems to be okay with just having labels. This PR's solution gives the user less text to process while still following W3C principles.  

# Checklist:

### DIFF
--- a/apps/website/src/routes/docs/headless/modal/examples/animation.tsx
+++ b/apps/website/src/routes/docs/headless/modal/examples/animation.tsx
@@ -32,7 +32,7 @@ export default component$(() => {
           <p class="mb-4 leading-5">
             You can update your profile here. Hit the save button when finished.
           </p>
-          <fieldset class="mb-1 flex items-baseline justify-between">
+          <div class="mb-1 flex items-baseline justify-between">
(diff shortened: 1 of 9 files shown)


thejackshelton:
Looks good. Merging.

PR #547, opened 2023-12-08
refactor(notes): change the Note system to reflect what notes are usually used for

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

I think this new note system would make more sense. Also, replaced the colors with the background/foreground CSS vars. Check it out, I can't see any color difference!

# Checklist:

### DIFF
--- a/apps/website/src/components/note/note.tsx
+++ b/apps/website/src/components/note/note.tsx
@@ -1,9 +1,8 @@
 import { component$, Slot, QwikIntrinsicElements } from '@builder.io/qwik';
 
 export enum NoteStatus {
-  Info = 'info',
-  Warning = 'warning',
-  Success = 'success',
+  Important = 'important',
+  FunFact = 'fun fact',
   Caution = 'caution',
 }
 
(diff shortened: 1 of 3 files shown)


thejackshelton:
The naming does not feel intuitive to me at first glance. I'm sure we'd get used to it, but the statuses are likely more expected, similar to how a programmer would expect an API with index in its name to start at the index of 0.

We can refactor success out.

`caution` or `destructive` here should be used very sparingly. We should continue using `warn` at the introduction, like melt ui does here:
https://www.melt-ui.com/docs/introduction

A destructive action is more concerning, and draw attention away from the introduction.

Where I do think a destructive action makes sense is below. Because if the user decides to use both flip and autoPlacement at the same time, it will break their component. I've purposefully listed this at the bottom, but it has an info tag on it currently.

image

We need a warning status, as there's a difference between something that people should be aware of, and things that are critical.

maiieul:
Hmmmm, yes, that's trickier than I thought.

This is not a big deal, we can keep it as it was. Didn't spent that much time on this refactor :upside_down_face: 

Just on a side note:

Melt-UI's "warning" is just their base "blockquote". It's just that their theme happens to be yellow.

I don't see much difference between "warning" and "caution".


Perhaps it would be best to just have Info and Warning? :grimacing: 

Although I kind of liked the "Fun Fact" type of note, could be with a ":thinking:" emoji kind of icon.

Also not sure that yellow warning looks very good. Just a warning icon would probably look better.

maiieul:
Feel free to close if you think it's not worth the trouble :+1:

maiieul:
Yeah let's just close this. You can reopen if you want :tongue:

thejackshelton:
I do think a fun fact note wouldn't hurt 😄 . Just that maybe it should be its own status rather than replacing info or warning.

PR #548, opened 2023-12-09
refactor(select-listbox): removed useVisibleTask

TheMcnafaha:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
A suprisingly quick code change gets rid of  the visibleTask

 your screenshots here -->

# Checklist:

### DIFF
--- a/packages/kit-headless/src/components/select/select-listbox.tsx
+++ b/packages/kit-headless/src/components/select/select-listbox.tsx
@@ -4,7 +4,6 @@ import {
   Slot,
   useContext,
   useSignal,
-  useVisibleTask$,
 } from '@builder.io/qwik';
 import SelectContextId from './select-context-id';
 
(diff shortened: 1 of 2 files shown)


thejackshelton:
Looks great! One thing I noticed, the keyboard behavior does not work unless something inside of the listbox is focused.

Is that the intended behavior? 

My assumption would be that it would "highlight" (not focus) something, regardless of if there is a focused item inside the select, although with that said, maybe there's something I'm missing.

Perhaps we don't want it affecting other components on our page. I'm still not sure of this behavior.

I do think it's much more fitting having this behavior in onKeyDown$. This goes with the principles of the framework 🎊 

preventDefault without a visible task is coming very soon. I got to try it last night and can confirm it works in preview at least, and I'll bet it might need to be used here somewhere. In particular the up and down arrow keys affecting the page scroll.

The upcoming way to preventDefault something in Qwik is like this:
```tsx
onKeyDown$={$sync((e: KeyboardEvent) => {
  if (e.key === "ArrowDown") {
    e.preventDefault();
  }
})}
```

Merging this PR. If this is not the intended behavior that I mentioned above, we can make a new one.

PR #553, opened 2023-12-12
refactor(headless/examples): changes the color scheme from css vars t…

TheMcnafaha:
draft pr. Meant to check my regex doesn't break CI.


TheMcnafaha:
@maiieul  has taken over this problem

PR #562, opened 2023-12-14
fix(auto-imports): remove pnpm workspaces to re-enable vscode auto-imports

maiieul:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Seems that using pnpm workspaces inside of nx is causing issues with vscode auto-imports. Perhaps @dmitry-stepanenko could confirm?

# Checklist:

### DIFF
--- a/packages/kit-headless/package.json
+++ b/packages/kit-headless/package.json
@@ -25,7 +25,6 @@
     "@builder.io/qwik": ">=1.2.19"
   },
   "dependencies": {
-    "@qwik-ui/utils": "workspace:*",
     "@floating-ui/dom": "^1.0.10",
     "body-scroll-lock-upgrade": "1.1.0",
     "country-list-json": "^1.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,8 +4,21 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-importers:
+dependencies:
+  '@fontsource-variable/inter':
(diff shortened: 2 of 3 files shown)


maiieul:
Looks like tailwind:lint is failing because of that change :thinking:

shairez:
This is not complete as we need to 

1. Move all of the dependencies inside of each project's package.json into the root package.json
2. Verify that all of the packages are being built correctly (also the website)
3. Find an automatic script that will allow us to sync the packages inside of each project with the root version numbers

maiieul:
Hmmm I see.

I'm not sure I'd be very productive at finding the script for maintaining package versions. 

Although the design kits don't need package versioning since it's copy&paste, and we wouldn't need to have @qwik-ui/utils if we used the cva & clsx libraries (so no package versioning here either). 

For me only headless is necessary as a versioned package as of now. If you can agree with me to use the cva & clsx libraries, then it simplifies the problem. https://discord.com/channels/990511757091033108/1144897805903724564/1184777208187867136

maiieul:
So I managed to "fix" this by removing old tailwind-kit code.

This is a good step towards a clean repo.

I'd say let's merge this and figure out later the script for package versioning.

(fluffy button is the same as the other PR I just opened, used it in here so that preview wouldn't complain about type errors)

maiieul:
Ups, I pushed a rebase and removed all the changes... I'm so clumsy with git x). Closing for now. I'll make another PR.

PR #564, opened 2023-12-14
RFC feat(aspect-ratio): add aspect-ratio component inspired from radix

maiieul:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

API is inspired by Radix. I guess they don't have this component for nothing :sweat_smile: . Probably useful to keep the right aspect-ratio in certain situations :stuck_out_tongue: 

I didn't make any test for it, but I guess this can (mostly) be copied from radix as well.

Would also require writing some docs, but I'd like to know what you guys think first.

# Checklist:

### DIFF
--- a/packages/kit-headless/src/components/aspect-ratio/aspect-ratio.tsx
+++ b/packages/kit-headless/src/components/aspect-ratio/aspect-ratio.tsx
@@ -1,0 +1,31 @@
+import { QwikIntrinsicElements, component$, Slot } from '@builder.io/qwik';
+
+type aspectRatioProps = QwikIntrinsicElements['div'] & { ratio?: number };
+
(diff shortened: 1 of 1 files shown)


thejackshelton:
Hm... yeah would definitely need docs. What is the use case for this component? There's a CSS `aspect-ratio` property.

It makes me wonder if this component existed when the aspect-ratio property did not exist.

I don't see other headless component libraries in the Svelte or Solid ecosystem that have an aspect-ratio component, and what that tells me is that we may not need this.

maiieul:
Yes, interesting, perhaps it's to enable accessibility OOTB?

maiieul:
Looks to me that the css `aspect-ratio` property would be even more powerful than the AspectRatio component (see MDN docs). Closing for now.

PR #565, opened 2023-12-15
Replace css vars with tw slate for all accordion examples

TheMcnafaha:
# What is it?

- [x] Feature / enhancement

# Description

Makes code more "copy & paste" friendly by changing all the css-vars to tw slates. 

# Screenshots/Demo
New version (most drastic example): 
new-version
Current version:
current-version

# Checklist:

### DIFF
--- a/apps/website/src/routes/docs/headless/accordion/examples/custom-heading.tsx
+++ b/apps/website/src/routes/docs/headless/accordion/examples/custom-heading.tsx
@@ -11,10 +11,10 @@ export default component$(() => {
   return (
     <>
       <div class="flex w-full justify-center">
-        <AccordionRoot class="w-[min(400px,_100%)]">
-          <AccordionItem class="border-b">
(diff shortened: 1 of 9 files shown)


maiieul:
We need to add a slate background on light-mode as well to the Showcase components (but keep it white for the design kits)

That would also allow us to not modify the padding

And no need for text-slate-50 imo

TheMcnafaha:
> And no need for text-slate-50 imo

You'd still need text-slate-50 because the consumer might have a different text color that would make our text invisible. 

Say they have text-slate-700 for the document body, if we remove our text-slate-50, they would not see the accordion trigger text (it will be the same color as the background). 

>We need to add a slate background on light-mode as well to the Showcase components (but keep it white for the design kits)

I'm confused here. This would require a change outside the accordion examples, right? 

> That would also allow us to not modify the padding

The consumer might still want the padding since they won't have access to our container.

Issue #517, opened 2023-11-15
Finalize the Select


shairez:

TheMcnafaha:
Roadmap, in order of how I want to progress:


- [ ] Make component use visual focus instead of normal focus were the W3 spec says so (affects most of current behavior)

Add missing keyboard support for popup (keys left listed below, listed under the same bullet point is behavior is related)
- [ ] Home & End
- [ ]  Tab & Alt+Up Arrow
- [ ]  PageUp & PageDown

PR #589, opened 2024-01-10
refactor(select): replaced focus behavior for visual focus (aria-sele…

TheMcnafaha:
The purpose of this draft pr is to finish the first bullet point on #517. This is likely going to be a big pr since I will likely need to refactor a lot. 

However, the  final pr will take whatever changes are done here and make them appear in a much smaller manner. Probably will result in splitting this draft pr into multiple prs. 

# What is it?

- [x] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Current select uses '.focus' to tell the user what option on the popup is being selected. It also doesn't implement 'aria-selected' properly. This pr should fix both. 

# Use cases and why

-1. Get closer to the W3 Spec


thejackshelton:
Stale and we have a new select version. Gonna close for now

PR #602, opened 2024-01-13
New bind:selectedIndex syntax, bug fixes

thejackshelton:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:

- [ ] I have ran `pnpm changeset` and documented my changes

### DIFF
--- a/.changeset/honest-timers-dream.md
+++ b/.changeset/honest-timers-dream.md
@@ -1,0 +1,5 @@
+---
+'@qwik-ui/headless': patch
+---
+
+bind syntax Sig suffix removed, new bind:selectedIndex prop, bug fixes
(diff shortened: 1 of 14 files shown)


.changeset/honest-timers-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik-ui/headless': patch
+---
+
+bind syntax Sig suffix removed, new bind:selectedIndex prop, bug fixes
```
shairez, line 5: I would separate type of changes into different commits, where each commit has it's own changeset

For example, one could be:

FEAT: (select) new `bind:selectedIndex` prop
Removed bind

Another could be

FIX: (select) now it's rendering correctly (or something like that)

PR #609, opened 2024-01-25
fix(dialog): ESC freezes browser

PatrickJS:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?
when you press Enter then ESC the dialog will be out of sync with the signal

### DIFF
--- a/packages/kit-headless/src/components/modal/modal.tsx
+++ b/packages/kit-headless/src/components/modal/modal.tsx
@@ -88,6 +88,11 @@ export const Modal = component$((props: ModalProps) => {
       {...props}
       role={props.alert === true ? 'alertdialog' : 'dialog'}
       ref={modalRefSig}
+      onKeyDown$={(event) => {
+        if (event.key === 'Escape') {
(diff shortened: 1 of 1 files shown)


thejackshelton:
Not sure I understand @PatrickJS, isn't the exact code already here?

```tsx
    const escapeKeyListener = overrideNativeDialogEscapeBehaviorWith(() => {
      showSig.value = false;
    });

    window.addEventListener('keydown', escapeKeyListener);
```

thejackshelton:
Is `showSig.value` not firing there? If so, let's refactor out `overrideNativeDialogEscapeBehaviorWith` entirely and use a sync$ to preventDefault, then keep the code here in this PR.

If you could add that commit.

PatrickJS:
@thejackshelton yeah the event fires from dialog and closes the modal then freezes the browser

PatrickJS:
@thejackshelton feel free to edit the code in the PR

PR #666, opened 2024-03-25
feat: add styled sonner draft

ditadi:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?
Increase the variety of components on qwik-ui by adding a sonner component from [qwik-sonner](https://github.com/diecodev/qwik-sonner).

https://github.com/qwikifiers/qwik-ui/assets/103508565/431b7457-9db3-43b1-953e-d46225449a28

# Checklist:


apps/website/src/_state/component-statuses.ts
```diff
@@ -27,6 +27,7 @@ export const statusByComponent: ComponentKitsStatuses = {
     RadioGroup: ComponentStatus.Draft,
     Separator: ComponentStatus.Beta,
     Skeleton: ComponentStatus.Beta,
+    Sonner: ComponentStatus.Draft,
```
thejackshelton, line 30: Perhaps naming this Toast? I know that Shadcn calls this sonner, but I believe that's because they already have a toast component.

Might be a bit easier to understand for consumers.

ditadi: I believe shadcn calls Sonner because the headless library was created with the name Sonner (https://sonner.emilkowal.ski/).

But, **I agree with you**, we can change to Toast and with this, we don't need to rush to have a headless Toast component from qwik-ui.

I will update.

ditadi: Changed.

thejackshelton: > I believe shadcn calls Sonner because the headless library was created with the name Sonner (https://sonner.emilkowal.ski/).
> 
> But, **I agree with you**, we can change to Toast and with this, we don't need to rush to have a headless Toast component from qwik-ui.
> 
> I will update.

Yeah I'd be curious to know if they'd still call it sonner if there wasn't a toast component they already had. I'll bet the didn't wanna call it Toast (2) haha.

maiieul:
Hmmm interesting I didn't comment in here.. 

We can't merge this as is cause the styled kit is expected to use headless components under the hood, and `cva`.

thejackshelton:
@maiieul @shairez can we close?

PR #720, opened 2024-04-24
feat: Add Chrome 113 testing to ensure backwards compatibility

cwoolum:
# What is it?

This PR add the ability to test with Chrome 113. I have tested it in Windows and Linux. 

### Linux

image

### Windows 

image

When running the command `nx run headless:e2e-chrome-113 --ui`, a script will run that downloads Chrome 113 for the appropriate OS the developer is using. 

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Checklist:


packages/kit-headless/package.json
```diff
@@ -10,6 +10,9 @@
   "module": "./index.qwik.mjs",
   "types": "./index.d.ts",
   "type": "module",
+  "scripts": {
+    "setup:chrome:113": "npx tsx ./scripts/prepare-chrome-113.ts"
```
cwoolum, line 14: Naming feels a bit verbose but there's no doubt what's going on here.

packages/kit-headless/playwright.config.ts
```diff
@@ -1,12 +1,34 @@
...
 
+const dirName = dirname(__filename);
+
+const currentPlatform = os.platform();
+
+let binaryPath;
+
+switch (currentPlatform) {
+  case 'darwin':
+    binaryPath = './browsers/chrome/113/chrome-darwin/Chromium.app';
```
cwoolum, line 20: Need someone with a Mac to test this.

packages/kit-headless/playwright.config.ts
```diff
@@ -41,6 +63,16 @@ export default defineConfig({
...
     },
 
+    {
+      name: 'popover-chrome-113',
+      use: {
+        launchOptions: {
+          executablePath: path.resolve(dirName, binaryPath),
+        },
+      },
+      grep: /popover/,
```
cwoolum, line 73: This grep isn't working. Need to look further into this.

packages/kit-headless/project.json
```diff
@@ -88,6 +88,21 @@
...
+    "setup-chrome-113": {
+      "executor": "nx:run-script",
+      "options": {
+        "script": "setup:chrome:113"
+      }
+    },
+    "e2e-chrome-113": {
+      "executor": "@nx/playwright:playwright",
+      "outputs": ["{workspaceRoot}/dist/.playwright/packages/kit-headless"],
+      "dependsOn": ["setup-chrome-113"],
```
cwoolum, line 100: Setup task always runs before the actual execution but bypasses if the folder already exists to improve performance.

cwoolum:
https://github.com/qwikifiers/qwik-ui-polyfill-browsers/pull/1 Needs to be merged before this one.

cwoolum:
@shairez @thejackshelton , this is ready for review.

packages/kit-headless/playwright.config.ts
shairez: @thejackshelton you have a mac now right? :)
did you have a chance to check this out?

packages/kit-headless/scripts/prepare-chrome-113.ts
```diff
@@ -0,0 +1,76 @@
...
+        console.log('Installing Chrome 113 for macOS 🍎');
+        break;
+
+      case 'win32':
+        console.log('Installing Chrome 113 for Windows 🪟');
+        break;
+    }
+
+    const downloadResponse = await fetch(
+      `https://github.com/cwoolum/qwik-ui-polyfill-browsers/raw/main/chrome/113/chrome-${os.platform()}.zip`,
```
shairez, line 38: I think this would be needed to change to `qwikifiers/`  right?

packages/kit-headless/package.json
shairez: I think it's good but let's use the dot notation to stay consistent with the other scripts
Plus, maybe we should even move this to the root package.json, as we have all of our scripts there anyway

WDYT?

shairez:
Great job @cwoolum ! thank you!

Left a few comments

packages/kit-headless/package.json
cwoolum: I changed it to be dot notation. I'm a bit torn on where it should live. I was on the fence about where to put the script. The script is very scoped to those tests so I didn't want to add complexity to the upper `package.json` if it really was this isolated. I think if we add another test package that requires chrome 113, we can refactor it up. The other nice benefit is that I can use it as a NX task dependency when it's in the same package

image

packages/kit-headless/scripts/prepare-chrome-113.ts
cwoolum: This should be resolved now

packages/kit-headless/playwright.config.ts
thejackshelton: Just checked this out. I got a downloading screen that went to 100% -> ran the tests, and running any test immediately fails.

@cwoolum does this mean the binary path is incorrect?

PR #721, opened 2024-04-24
chore(website vite.config): add  visible chunk names to rollupOptions…

maiieul:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [x] Other

# Why is it needed?

This allows seeing the chunk file names instead of `q-[hash].js` in preview or production:

image

This is a bit of a hack, but I think it will be helpful to have this, maybe up until v1, so that we have better visibility into what's happening when looking up the network tab.

I'll try to create a Qwik-City PR at some point to make this happen when setting `qwikVite.debug` to true.

# Checklist:


thejackshelton:
Looks good! Merging.

PR #734, opened 2024-05-11
New Modal API Changes

thejackshelton:
# What is it?

This PR proposes a couple of developer experience improvements to the modal.

In a previous release, the following components have been deprecated:

- ModalHeader
- ModalContent
- ModalFooter

These components were native header, div, and footer elements and did nothing special under the hood. We are deprecating them in order to simplify the API and make it more consistent with the rest of the components.

The new components are:

#### <Modal.Root>

This is the main container of the modal, and now holds the major props and configuration. Examples include:

- 'bind:show'?: Signal<boolean>;
- closeOnBackdropClick?: boolean;
- alert?: boolean;
- onShow$?: QRL<() => void>;
- onClose$?: QRL<() => void>;

#### <Modal.Panel>

Previously `<Modal />` the modal panel is the dialog element that is rendered on top of the page. Its props have since been moved to the `<Modal.Root />` component, please refer to the docs for more information.

#### <Modal.Trigger>

The modal now comes with a default trigger, which will open the modal when clicked.

#### <Modal.Title>

This computes the accessible name from the string children of the modal.

#### <Modal.Description>

This computes the accessible description from the string children of the modal.

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:


packages/kit-headless/src/components/modal/modal-trigger.tsx
```diff
@@ -0,0 +1,16 @@
...
+  const context = useContext(modalContextId);
+
+  const handleClick$ = $(() => {
+    context.showSig.value = !context.showSig.value;
+  });
+
+  return (
+    <button onClick$={[handleClick$, props.onClick$]} {...props}>
+      <Slot />
+    </button>
```
maiieul, line 14: Shouldn't the ModalTrigger button also have some aria attributes on it?

Like in Radix:

```tsx
const DialogTrigger = React.forwardRef<DialogTriggerElement, DialogTriggerProps>(
  (props: ScopedProps<DialogTriggerProps>, forwardedRef) => {
    const { __scopeDialog, ...triggerProps } = props;
    const context = useDialogContext(TRIGGER_NAME, __scopeDialog);
    const composedTriggerRef = useComposedRefs(forwardedRef, context.triggerRef);
    return (
      <Primitive.button
        type="button"
        aria-haspopup="dialog"
        aria-expanded={context.open}
        aria-controls={context.contentId}
        data-state={getState(context.open)}
        {...triggerProps}
        ref={composedTriggerRef}
        onClick={composeEventHandlers(props.onClick, context.onOpenToggle)}
      />
    );
  }
);
```

Also I feel like if we add a ModalTrigger we'll need to add an `asChild` prop to it. This way users can directly put their `<Button>` component in the ModalTrigger.

maiieul (commented):
WDYT about having a `Modal.Close`? Could be convenient in order to not have to do a `signal = false` manually.

apps/website/src/routes/docs/headless/modal/examples/nested-popover.tsx
```diff
@@ -1,17 +1,16 @@
...
 
 export default component$(() => {
   useStyles$(styles);
   const isOpen = useSignal(false);
 
   return (
-    <>
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 10: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/animatable.tsx
```diff
@@ -7,11 +7,11 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <div class="modal-container">
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root class="modal-container" bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/auto-focus.tsx
```diff
@@ -7,11 +7,11 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <div class="modal-container">
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root class="modal-container" bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/backdrop-close.tsx
```diff
@@ -7,11 +7,11 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <div class="modal-container">
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root class="modal-container" closeOnBackdropClick={false} bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/backdrop.tsx
```diff
@@ -7,11 +7,11 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <div class="modal-container">
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root class="modal-container" bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/bottom-sheet.tsx
```diff
@@ -7,30 +7,28 @@ export default component$(() => {
...
-          </Label>
-          <footer>
-            <button onClick$={() => (isOpen.value = false)}>Cancel</button>
-            <button onClick$={() => (isOpen.value = false)}>Save Changes</button>
-          </footer>
-        </Modal.Root>
-      </div>
-    </>
+    <Modal.Root class="modal-container" bind:show={isOpen}>
+      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: shouldn't this be Modal.Trigger?

apps/website/src/routes/docs/headless/modal/examples/focus-trap.tsx
```diff
@@ -7,16 +7,16 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <>
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

apps/website/src/routes/docs/headless/modal/examples/transition.tsx
```diff
@@ -7,11 +7,11 @@ export default component$(() => {
   const isOpen = useSignal(false);
 
   return (
-    <div class="modal-container">
-      <button class="modal-trigger" onClick$={() => (isOpen.value = true)}>
+    <Modal.Root class="modal-container" bind:show={isOpen}>
+      <Modal.Trigger class="modal-trigger" onClick$={() => (isOpen.value = true)}>
```
maiieul, line 11: onClick can be removed here no?

packages/kit-headless/src/components/modal/modal-trigger.tsx
thejackshelton: Yes, the following aria attributes should be added to the modal trigger, thanks for pointing that out.

As for the asChild pattern, I don't see the large benefit. The component needs to know its direct child, which goes against the philosophy of both slots and Qwik's out of order rendering.

To be consistent, then half the library would need an `asChild` prop, which lengthens the markup. I'm not sure having an inline component everywhere would be the best idea either.

I don't see the article you've linked. In either case, a trigger is more the most part a button. There are two reasons for the switch:

1. A better developer experience and a11y as a default.
2. Invokers. **https://open-ui.org/components/invokers.explainer/**

thejackshelton:
> WDYT about having a `Modal.Close`? Could be convenient in order to not have to do a `signal = false` manually.

 I saw it in Radix, but didn't want to add it in until I saw a particular need for it. After looking through the examples, I'll go ahead and add this to the PR, as I think it would improve the developer experience.

packages/kit-headless/src/components/modal/modal-trigger.tsx
maiieul: Lol sorry, forgot to link it: https://medium.com/@bryanmylee/aschild-in-react-svelte-vue-and-solid-for-render-delegation-645c73650ced

maiieul:
Hmmm I guess in any ase if we need so aria attributes on the `Modal.Trigger` then we need this component.

Also having a `Modal.Root`means we can simply do the `bind:show={isOpen}` on it and no need to play with the signals in the other components on a general basis.

As for the asChild prop, if it means we must create an inline component for it, then perhaps it's best to address that in a future PR.

So I guess LGTM 👍

PR #763, opened 2024-05-24
fix(cli): add try/catch block to create config file

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:

### DIFF
--- a/packages/cli/bin/index.ts
+++ b/packages/cli/bin/index.ts
@@ -193,16 +193,20 @@ async function handleInit() {
   });
 
   // CREATE CONFIG FILE
-  execSync(
-    `${
-      getPackageManagerCommand().exec
-    } nx g qwik-ui:init --interactive false --project-root=${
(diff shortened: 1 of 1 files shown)


TheMcnafaha:
It's unix tradition that cli programs that error out send a positive integer value to stdout or stderr. If this script is very close to the user, it probably makes sense to send the error message to stdout. If the script is further away, or it runs in the background, stderr is usually the go to.  

It appears that this is convention is already being used through this function:
```js
function exitWithError(message: string) {
  log.error(message);
  cancel();
  process.exit(1);
}
```
Do you think it would make sense to throw an error that writes to stdout if the config file is not written? And if not, what exactly is the point of a try catch that only sends a console log? If the error is worth catching I would think it's worth throwing too.

PR #818, opened 2024-06-02
Toc add nested uls

TheMcnafaha:
# What is it?

- [x] Feature / enhancement
- [x] Bug
- [x] Docs / tests
- [ ] Other

# Why is it needed?
Current TOC is not accessible. 

# Checklist:


maiieul:
@TheMcnafaha Thanks for your hard work! How did you end up with this RecursiveJSX function 😱 I've never seen that in Qwik 😅  Scary.. but quite impressive tbh!

Remarks:
- Why make a headless TOC component? The current code is not even headless is it? In any case, even if it is headless, I don't think we want to have a headless TOC component as of now. Could you simply put it back in apps/website/src/components/toc/toc.tsx?
- Can we get rid of the bullet point on the first item?
image
- There is no highlighting of the current section of the page anymore. If you go to qwik UI and click on a TOC item, you'll see that it remains highlighted. Also scrolling through the page will change the highlighting depending on where you are on the page.
- Honestly I'm a bit warry of doing such tricks as `const infiniteStopper = JSON.parse(JSON.stringify(props.headings));`. We already have lots of complexity in many places in Qwik UI. I'd rather avoid more if possible 🙏. The RecursiveJSX seems quite clever, but maybe it's too clever for just 2-3 levels of nesting 🤔 Perhaps we could make 2-3 separate components for each level of nesting, and simply loop over them in an inlined component 🤔

TheMcnafaha:
Thanks for your feedback @maiieul, it is of great detail. I'll try my best to answer to it. 

>Why make a headless TOC component? The current code is not even headless is it? In any case, even if it is headless, I don't think we want to have a headless TOC component as of now. Could you simply put it back in apps/website/src/components/toc/toc.tsx?

For two reasons: (1) I want to add tests and I don't know of a different way, and (2) I feel that making a TOC component is a good component to have, so I got us started down that path.

> Can we get rid of the bullet point on the first item?

I cannot see this. I'm guessing this is due to the browser. I'm on Firefox and see this:
image

I believe the same is happening with the highlighting since it seems to be working for me.

> Honestly I'm a bit warry of doing such tricks as const infiniteStopper = JSON.parse(JSON.stringify(props.headings));. We already have lots of complexity in many places in Qwik UI. I'd rather avoid more if possible 🙏. 

I agree. There is probably a more convenient way of doing this, but I this is a very readable solution and I believe it's good enough for a "secret" component to use. I'll do more research for deep cloning but I don't think there's a more readable solution. 

>The RecursiveJSX seems quite clever, but maybe it's too clever for just 2-3 levels of nesting 🤔 Perhaps we could make 2-3 separate components for each level of nesting, and simply loop over them in an inlined component 🤔

I think this is where we are having different views for the TOC. We both want to make it better but I want to do that by adding more headings; the 2-3 levels of nesting will become 2-4. 

As for the recursive part, I don't like using recursion. To me, it's a better concept than inheritance, but still a general thing to avoid.
However, the one use case that recursion shines one is when you are dealing with tree-like structures. It makes the children part of trees easy to handle once you get over the recursive bump. I have heard it's possible to use loops to achieve the same result at the cost of readability and code complexity. 

 I don't think I would be interested in hard-coding an inline component that handles 2-3 nesting levels for three reasons: (1) as I said before, I want to add more levels, (2) I think it will be quite a challenge to do, and (3) it keeps us away from making a general TOC component.

thejackshelton:
@TheMcnafaha thanks for taking the initiative on this and the hard work.

If there's headless logic (more than just styles) in a table of contents component, then it makes sense to have as a headless component.

The only place I have found a table of contents component so far in a headless library, has been Melt UI. 

As for the nesting structure, I think (for the docs) we want whatever is most navigable. What do other popular documentation sites have? Do they have more than 2-3 nested headings?

As a general rule, we don't want to add functionality to a component unless it's very clear that there's a demand for it. Do other headless libraries offer this level of nesting? Are there any good use cases for this? Maybe an interactive textbook?

@TheMcnafaha happy to pair on this sometime if you'd like. If we could avoid the visible task somehow that would be awesome.

maiieul:
Currently no single page has more than 2 levels of nesting. I'm thinking we might have some exceptions with a third level, but 4 seems like way too much imo. I think let's first determine together what kind of TOC we want, and then see if we need to change anything on the current implementation. Btw the current one is almost a clone of the shadcn one, so I guess it's not that problematic 😋

What needs to be determined imo:
- Do we want to create/maintain a headless TOC component? (I'd say maybe in 1-2 years, not now)
- How many levels of nesting do we want to keep a good UX? (I'd say 2... 3 seems even like too much)
- Do we want to keep the highlighting? (definitely yes)
- Do we really need test for a self-owned component we're not going to change much? (I don't think it's necessary)
- Is <ul> nesting that important that's it's worth the effort now? (I'm not sure 🙈)

TheMcnafaha:
> the current one is almost a clone of the shadcn one

Yes and no. I can see the similarity but their tree function also uses recursion, they just use a ternary check instead of splitting the template into two recursive cases: 
```js
function Tree({ tree, level = 1, activeItem }: TreeProps) {
  return tree?.items?.length && level < 3 ? (
    <ul className={cn("m-0 list-none", { "pl-4": level !== 1 })}>
      {tree.items.map((item, index) => {
        return (
          <li key={index} className={cn("mt-0 pt-2")}>
            <a
              href={item.url}
              className={cn(
                "inline-block no-underline transition-colors hover:text-foreground",
                item.url === `#${activeItem}`
                  ? "font-medium text-foreground"
                  : "text-muted-foreground"
              )}
            >
              {item.title}
            </a>
// in my code, this is called "recursive_nested" but they just user a ternary here
            {item.items?.length ? (
// recursive part below:
              <Tree tree={item} level={level + 1} activeItem={activeItem} />
            ) : null}
          </li>
        )
      })}
    </ul>
  ) : null
}
```
This key component is missing from our implementation, meaning that instead of having nested uls we have one row of lis. 

> How many levels of nesting do we want to keep a good UX? (I'd say 2... 3 seems even like too much)

The title counts as a nesting level. It wouldn't visually change that much, it would just add semantic meaning to readers. 

> Do we want to keep the highlighting? (definitely yes)

I agree. I think I did a bad job explaining myself, so let me try again: it works on firefox, me browser, so  could you please tell me browser are you using so that I can know what to test on. 

>Is <ul> nesting that important that's it's worth the effort now? (I'm not sure 🙈)

If you want to make our TOC accesssible, it's a must have. Screen readers do not take padding into account. I know that the chance that a developer-centric site is targeted by a screen reader is not very likely, but I don't want to contribute to a headless library with an inaccessible TOC.  It would actually be really funny if a headless UI library had missing accessibility of a core component if it wasn't the same project I was backing. 

There's a reason why every other TOC has nested uls, but maybe we need the opinion of a more experienced developer. I don't use twitter, but maybe one of you guys can ask that.

TheMcnafaha:
@thejackshelton I'll accept the pairing invitation. Maybe afterwards we could also discuss the qwik core docs refactor.

thejackshelton:
> > the current one is almost a clone of the shadcn one
> 
> Yes and no. I can see the similarity but their tree function also uses recursion, they just use a ternary check instead of splitting the template into two recursive cases:
> 
> ```js
> function Tree({ tree, level = 1, activeItem }: TreeProps) {
>   return tree?.items?.length && level < 3 ? (
>     <ul className={cn("m-0 list-none", { "pl-4": level !== 1 })}>
>       {tree.items.map((item, index) => {
>         return (
>           <li key={index} className={cn("mt-0 pt-2")}>
>             <a
>               href={item.url}
>               className={cn(
>                 "inline-block no-underline transition-colors hover:text-foreground",
>                 item.url === `#${activeItem}`
>                   ? "font-medium text-foreground"
>                   : "text-muted-foreground"
>               )}
>             >
>               {item.title}
>             </a>
> // in my code, this is called "recursive_nested" but they just user a ternary here
>             {item.items?.length ? (
> // recursive part below:
>               <Tree tree={item} level={level + 1} activeItem={activeItem} />
>             ) : null}
>           </li>
>         )
>       })}
>     </ul>
>   ) : null
> }
> ```
> 
> This key component is missing from our implementation, meaning that instead of having nested uls we have one row of lis.
> 
> > How many levels of nesting do we want to keep a good UX? (I'd say 2... 3 seems even like too much)
> 
> The title counts as a nesting level. It wouldn't visually change that much, it would just add semantic meaning to readers.
> 
> > Do we want to keep the highlighting? (definitely yes)
> 
> I agree. I think I did a bad job explaining myself, so let me try again: it works on firefox, me browser, so could you please tell me what browser are you using so that I can know what to test on.
> 
> > Is  nesting that important that's it's worth the effort now? (I'm not sure 🙈)
> 
> If you want to make our TOC accesssible, it's a must have. Screen readers do not take padding into account. I know that the chance that a developer-centric site is targeted by a screen reader is not very likely, but I don't want to contribute to a headless library with an inaccessible TOC. It would actually be really funny if a headless UI library had missing accessibility of a core component if it wasn't the same project I was backing.
> 
> There's a reason why every other TOC has nested uls, but maybe we need the opinion of a more experienced developer. I don't use twitter, but maybe one of you guys can ask that.

Sounds good to me!

thejackshelton:
> @thejackshelton I'll accept the pairing invitation. Maybe afterwards we could also discuss the qwik core docs refactor.

Awesome. I'll reach out to you in discord

apps/website/src/components/toc/toc.tsx
```diff
@@ -1,33 +1,138 @@
...
 
 export const DashboardTableOfContents = component$(
   ({ headings }: { headings: ContentHeading[] }) => {
-    const itemIds = headings.map((item) => item.id);
-    const activeHeading = useActiveItem(itemIds);
-
     if (headings.length === 0) {
       return null;
     }
+    console.log(headings);
```
maiieul, line 15: can you remove this log pls? 🙏

apps/website/src/components/toc/toc.tsx
```diff
@@ -60,45 +165,35 @@ const useActiveItem = (itemIds: string[]) => {
...
-              )}
-            >
-              {heading.text}
-            </a>
-          </li>
-        );
-      })}
-    </ul>
-  ) : null;
+export const AnchorThing = component$<AnchorThingProps>((props) => {
```
maiieul, line 172: AnchorThing? Maybe just `Anchor`?

apps/website/src/components/toc/toc.tsx
```diff
@@ -1,33 +1,138 @@
...
+import { cn } from '@qwik-ui/utils';
+type TableOfContentProps = { headings: ContentHeading[] };
+interface Node extends ContentHeading {
+  level: number;
+  children: Array<Node>;
+  activeItem: string;
+}
+type Tree = Array<Node>;
+export const TableOfContent = component$<TableOfContentProps>((props) => {
+  const infiniteStopper = JSON.parse(JSON.stringify(props.headings));
```
maiieul, line 37: I know I'm being picky, but would it be possible to get rid of this `JSON.parse(JSON.stringify(props.headings))` trick? I'm a huge fan of KISS and I'm a bit afraid this is a bit over-engineering an internal component that will be hard to debug in the future. I mean I'm not 100% against merging this but if there's a bug in the future, we might have to bring back the old TOC cause it's way more simple ^^

Maybe doing it more like shadcn? I thought what I did was exactly shadcn but you're right I missed something 😅 .. But maybe if we limit the logic to three levels of nesting then we don't need that trick somehow?

apps/website/src/components/toc/toc.tsx
TheMcnafaha: Yes lol

apps/website/src/components/toc/toc.tsx
TheMcnafaha: :+1:

apps/website/src/components/toc/toc.tsx
TheMcnafaha: You cannot modify any of the heading objects. If you do you cause  an infinite loop, which is really fun to debug when you're using recursive code. I could just write a loop that maps over the objects instead. Do you think that would be an improvement? 

> I'm a huge fan of KISS and I'm a bit afraid this is a bit over-engineering an internal component that will be hard to debug in the future

I don't think a looping non-recursive alternative would be any easier to debug. I can do it if that's what it would take to have your blessing though.  

> Maybe doing it more like shadcn?

I'm 99% sure we are doing it like shadcn. In fact, I think we could fork their recursive function and it would just work.

maiieul: If you can reproduce the shadcn version of it 100% without spending a week on it, this would be much more readable no? Can you give it a try?

> I could just write a loop that maps over the objects instead. Do you think that would be an improvement?
I'm not sure I understand the difference, but sounds simpler than the current impl 🙈

TheMcnafaha: > If you can reproduce the shadcn version of it 100% without spending a week on it, this would be much more readable no? Can you give it a try?

The shadcn uses a lot of ternaries. 
```js
return tree?.items?.length && level < 3 ?
```

```js
      {item.items?.length ? (
              <Tree tree={item} level={level + 1} activeItem={activeItem} />
            ) : null}
          </li>
```
I'm personally not a fan, but I need this part merged before going any further with the TOC so i'll do it.

maiieul: Thank you 🙏

TheMcnafaha: Should all be done. If it all looks good, I'll rebase main so we can merge.

maiieul (approved):

apps/website/src/components/toc/toc.tsx
maiieul: LGTM!

maiieul:
Thanks again @TheMcnafaha! I've been quite picky but really cool PR 🔥   

Side note, I'm not sure of the implications of using that recursive pattern in Qwik, so I'm squashing/merging in case we get any issues.

PR #827, opened 2024-06-06
fix: get all popover tests running with polyfill

cwoolum:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

Chromium 109-113 did not properly support the popover but reported that they did. This led to the polyfill not activating which caused our E2E tests to break. We are dropping down to Chromium 108 to validate the polyfill as users of Chrome would see it before the popover API was supported.

# Checklist:


thejackshelton (approved):
LGTM! 🙌 

Merging

Side note: @cwoolum I think you can use .toHaveCSS instead of the .evalute method from playwright

PR #854, opened 2024-06-15
docs(toc): improve code readability

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:

### DIFF
--- a/apps/website/src/components/toc/toc.tsx
+++ b/apps/website/src/components/toc/toc.tsx
@@ -10,49 +10,37 @@ export const DashboardTableOfContents = component$(
     return (
       <div class="space-y-2">
         <div class="font-medium">On This Page</div>
-        <TableOfContent headings={headings} />
+        <TableOfContents headings={headings} />
       </div>
     );
   },
 );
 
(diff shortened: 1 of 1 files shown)


TheMcnafaha:
Also, there's a bug that was introduced from me changing the template to match shadcn's. The h1 is no longer being rendered since it's the starting point and the recursion only adds child nodes,  so no h1.

TheMcnafaha:
Also x2, some of the mdx doesn't have continuous heading levels. Separator is a good example, so the error is being thrown.

maiieul:
> Also x2, some of the mdx doesn't have continuous heading levels. Separator is a good example, so the error is being thrown.

I fixed this in another PR. 

> Also, there's a bug that was introduced from me changing the template to match shadcn's. The h1 is no longer being rendered since it's the starting point and the recursion only adds child nodes, so no h1.

Can you fix this in another PR?

TheMcnafaha:
> Can you fix this in another PR?
Sure. I'll fork this branch and ping you ind discord once i've fixed it.

PR #858, opened 2024-06-17
Toc readability

TheMcnafaha:
The more I tried to fix the template logic  the more confident I have become that the shadcn nested ternaries JSX logic is harder to read and way more error prone. 

This is a workaround. I don't want to fix this bug unless we change the template to multiple ifs statements with returns.

### DIFF
--- a/apps/website/src/components/toc/toc.tsx
+++ b/apps/website/src/components/toc/toc.tsx
@@ -29,7 +29,8 @@ const TableOfContents = component$<TableOfContentsProps>(({ headings }) => {
   const itemIds = headings.map(({ id }) => id);
   const activeHeading = useActiveItem(itemIds);
   const tree = buildTree(sanitizedHeadings);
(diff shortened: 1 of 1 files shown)


maiieul:
@TheMcnafaha I merged your upwards discontinuity fix PR https://github.com/qwikifiers/qwik-ui/pull/866

Is this one ready for review? You might have to merge changes from latest main first.

TheMcnafaha:
@maiieul I just rebased main. Should be good to merge! :smile:

TheMcnafaha:
I did undo one of the changes. I reverted the change from minimizing the switch statements from one word to what they originally were (eg "down" => "down one level").

Now that we have nth levels differences due to upwards discontinuity, I think we should make it explicit when we have a one-level difference.

TheMcnafaha:
Also, as i went over the code, I managed to see that there's one potential bug in it. 

It has to do with how we handle upwards discontinuities. The fix is simple but I want to wait until this branch is merged to avoid merge conflicts.

PR #862, opened 2024-06-19
fix(styled-kit): add type keyword to imported types

zougari47:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [x] Other

# Why is it needed?

**NOTE**: THIS IS FOR STYLED KIT
I was trying to copy/paste the textarea component in the styled kit in Qwik UI and I get this error:
```PropsOf' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.ts(1484)```

I was hesitating to do this PR because, in this code base I didn't get the error. (I think something in the ts config and I'm eager to know it). But the Card and Input components have the `type` keywords and others do not.
So it should be either in all or in none. What do you think?

# Checklist:

### DIFF
--- a/apps/website/src/routes/docs/styled/accordion/index.mdx
+++ b/apps/website/src/routes/docs/styled/accordion/index.mdx
@@ -21,7 +21,7 @@ qwik-ui add accordion
 ```
 
 ```tsx
-import { component$, Slot, PropsOf } from '@builder.io/qwik';
+import { component$, Slot, type PropsOf } from '@builder.io/qwik';
 
 import { Accordion as HeadlessAccordion } from '@qwik-ui/headless';
 import { cn } from '@qwik-ui/utils';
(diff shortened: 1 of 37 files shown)


zougari47:

maiieul:
Hmmm thanks @zougari47 for pointing this out! Since most of Qwik projects do indeed require the type keyword I think we should indeed use this syntax by default. Besides, projects that don't use it probably won't be affected by having the keyword, or will straight up remove it on save.

Thanks again 🙏

zougari47:
> Hmmm thanks @zougari47 for pointing this out! Since most of Qwik projects do indeed require the type keyword I think we should indeed use this syntax by default. Besides, projects that don't use it probably won't be affected by having the keyword, or will straight up remove it on save.
> 
> Thanks again 🙏

I'm glad it helped. 
I would like to know why In this project didn't give an error(out of curiosity).

PR #866, opened 2024-06-23
fix(toc): handles upwards discontinuity

TheMcnafaha:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:


TheMcnafaha:
Keep in mind that there's still one bug left: 
If you have two h1s, it stops rendering anything the it sees the second h1. 

How to fix that would require throwing another error or allowing multiple h1s.

PR #868, opened 2024-06-26
refactor: website toc perf

thejackshelton:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:


TheMcnafaha:
Impressive work @thejackshelton !

My only question are the file changes on `apps/website/src/routes/docs/headless/dropdown`. Is this a a purposeful change or something that slipped up? It don't see how they relate to toc performance. I don't think this is a major problem, but I do think it would be nice to give them their own commit: "feat(dropdown): uses qwik-only text area"

thejackshelton:
> Impressive work @thejackshelton !
> 
> My only question are the file changes on `apps/website/src/routes/docs/headless/dropdown`. Is this a a purposeful change or something that slipped up? It don't see how they relate to toc performance. I don't think this is a major problem, but I do think it would be nice to give them their own commit: "feat(dropdown): uses qwik-only text area"

It's a mistake I saw and was too lazy to make a new PR for a one line change haha

TheMcnafaha:
Happens to the best of us. If you want, I could make the additional commit.

thejackshelton:
> Happens to the best of us. If you want, I could make the additional commit.

That'd be awesome :)

TheMcnafaha:
While ensuring that my rebase didn't break the code, I realized there's a minor problem with how the scroll bar is handled:
image

Should this prevent merging?

thejackshelton-kunaico:
> While ensuring that my rebase didn't break the code, I realized there's a minor problem with how the scroll bar is handled (on firefox 127): image This is was not caused by my rebasing (it's present in the current state of the branch)
> 
> Should this prevent merging?

Yeah the firefox scrollbars end up looking a bit smaller compared to the chrome ones. Perhaps we're missing a vendor prefix somewhere.

thejackshelton:
Yeah the firefox scrollbars end up looking a bit smaller compared to the chrome ones. Perhaps we're missing a vendor prefix somewhere.

Either way I don't think it's a blocker. Firefox by default has smaller scrollbars, so that is expected of someone using firefox. It is also a smaller fraction of the browser usage.

I don't think a small scrollbar on firefox blocks the removal of a whole bunch of dependencies we aren't really using.

We can improve this when we get our own scroll area component

maiieul:
I'm merging this cause it works independently of the react deps. I'll try to remove them in a separate PR and see if we get the same pnpm.test cli error.

Issue #853, opened 2024-06-15
Finish Tooltip to a beta stage


cwoolum:
## Required Features for Beta

TBD

## Existing Tooltip Libraries Comparison

To determine a basic feature set for the tooltip component, I compared the tooltop components from Melt, Kobalte, and Radix. I'll start with the common features they all share. You can read about them more individually below.

### Common Features
---------------

#### Trigger Mechanism

-   Focus or Hover: All frameworks open the tooltip when the trigger element is focused or hovered.

-   Melt UI: Opens on focus or hover.

-   Kobalte UI: Opens on keyboard focus or mouse hover.

-   Radix UI: Opens on focus or hover.

#### Closing Mechanism

-   Activation or Escape Key: All frameworks close the tooltip when the trigger element is activated or when the Escape key is pressed.

-   Melt UI: Closes when the trigger is activated or Escape is pressed.

-   Kobalte UI: Closes when the trigger is activated or Escape is pressed.

-   Radix UI: Closes when the trigger is activated or Escape is pressed.

#### Customizable Delays

-   Open and Close Delays: All frameworks support customizable delays for opening and closing the tooltip.

-   Melt UI: openDelay and closeDelay props.

-   Kobalte UI: openDelay and closeDelay props.

-   Radix UI: delayDuration and skipDelayDuration props (via Provider) and delayDuration prop on Root.

#### Optional Arrow

-   Tooltip Arrow: All frameworks provide an optional arrow component that points to the trigger element.

-   Melt UI: Optional arrow component.

-   Kobalte UI: Optional Tooltip.Arrow component.

-   Radix UI: Optional Tooltip.Arrow component.

#### Portaling

-   Portal Support: All frameworks support rendering the tooltip content into a portal, typically the body element, to manage overlay issues.

-   Melt UI: portal prop.

-   Kobalte UI: Tooltip.Portal component.

-   Radix UI: Tooltip.Portal component.

#### Positioning

- Melt UI:  Provides positioning via the `positioning` prop.

- Kobalte UI: Provides positioning via the `placement` prop and `gutter` props. Also supports `flip`.

- Radix UI: Provides positioning using `side` and `sideOffset` props.

#### Accessibility

-   ARIA and Keyboard Support: All frameworks follow the ARIA design pattern for tooltips and provide proper keyboard interactions for accessibility.

[...]

PR #882, opened 2024-07-11
feat(tooltip): implement tooltip to beta phase

cwoolum:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests
- [ ] Other

# Why is it needed?

Implements #853 

# Checklist:


packages/kit-headless/src/components/tooltip/tooltip-panel.tsx
```diff
@@ -0,0 +1,23 @@
...
+ */
+export const HTooltipPanel = component$((props: HTooltipPanelProps) => {
+  const context = useContext(TooltipContextId);
+
+  return (
+    <HPopoverPanel
+      {...props}
+      role="tooltip"
+      id={context.localId}
+      data-state={context.state.value}
```
cwoolum, line 18: There are 4 possible states for this component. Do we have individual `data-` fields for each or a single state field?

packages/kit-headless/src/components/tooltip/tooltip-root.tsx
```diff
@@ -1,12 +1,96 @@
...
+ * HTooltipRoot is the root component for the Tooltip.
+ */
+export const HTooltipRoot = component$((props: TooltipProps) => {
+  const {
+    placement = 'top',
+    id,
+    gutter,
+    delayDuration = 0,
+    flip,
+    onOpenChange$: _,
```
cwoolum, line 61: Need to implement events still.

packages/kit-headless/src/components/tooltip/tooltip-root.tsx
```diff
@@ -1,12 +1,96 @@
...
 
   return (
-    <HPopoverRoot hover={hover} floating={floating} {...rest}>
-      <Slot />
+    <HPopoverRoot
+      manual
+      hover={true}
+      bind:anchor={triggerRef}
+      floating={placement}
+      id={localId}
```
cwoolum, line 87: Where should `localId` actually go?

packages/kit-headless/src/components/tooltip/tooltip-root.tsx
thejackshelton: Hm... this is an interesting one because the popover actually has an event called `onToggle$`. Although that is not consistent with what we have in the library. I wonder if we can just gain that functionality whenever someone passes `onOpenChange$`.

packages/kit-headless/src/components/tooltip/tooltip-root.tsx
thejackshelton: localId is the "same id" across the headless component. it's usually for when you have 1:1 components. Such as one trigger, one input, one root, etc.

Then you can prefix these in any file with a new variable:

```
const inputId = `${context.localId}-input`
```

And the component knows you are referring to that instances input.

packages/kit-headless/src/components/tooltip/tooltip-panel.tsx
thejackshelton: answered in DM. happy to paste here if appropriate

cwoolum:
@thejackshelton , implemented `onOpenChange` and removed animations for the docs. Let me know what you think.

packages/kit-headless/src/components/tooltip/tooltip.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    await d.getTrigger().press('Escape');
+    await expect(d.getTooltip()).not.toBeVisible();
+  });
+});
+
+test.describe('Placement and Positioning', () => {
+  const placements = ['top', 'right', 'bottom', 'left'] as const;
+
+  test.describe('Tooltip Placements', () => {
+    placements.forEach((placement) => {
```
shairez, line 80: sometimes generated tests from a loop are harder to debug

An alternative approach is to encapsulate all of the setup in a dedicated setup function for this test (for example, everything until line 92 and just create this test 4 time and call the "custom setup" function with a param for the placement. 

It's not a must, but it might be easier to debug if we'll run into issues in the future.

shairez:
Great job @cwoolum ! 👏

Left one comment, other than that, LGTM! 🙏

packages/kit-headless/src/components/tooltip/tooltip.test.ts
cwoolum: I refactored some helpers into the driver split out the different cases.

thejackshelton:
closing #882 in favor of #934 which builds upon this PR

PR #884, opened 2024-07-13
fix: breadcrumb class override instead of concatenate

zougari47:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

Because we spread props after the class attribute, It gets overridden instead of concatenated with the `cn` function.
I have checked other components and it seems this problem is only in the Breadcrumb component.

# Checklist:

- [ ] I have ran `pnpm change` and documented my changes (I don't konw if this should be patched or not !!)


maiieul (approved):
oh thanks! Good catch @zougari47! And sorry for the oversight 😬

PR #906, opened 2024-07-26
single shiki instance & on the server w/ dev

thejackshelton:
# What is it?

Preview route transitions are now much faster, shiki should now be working correctly in dev mode and preview, as I believe the previous error was due to multiple instances being created each time an example was rendered.

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:

### DIFF
--- a/apps/website/src/components/highlight/highlight.tsx
+++ b/apps/website/src/components/highlight/highlight.tsx
@@ -1,25 +1,23 @@
-import {
-  ClassList,
-  PropsOf,
-  component$,
-  useSignal,
-  useTask$,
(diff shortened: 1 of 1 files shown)


maiieul (approved):
Cool find! I didn't think of creating the highlighter and then awaiting it in the component 🙈 

Just re-added the track(() => code) as it is required for the MakeItYours's CopyCSSConfig. There's still a flicker. Maybe we can fix it with a computed sig, but for another PR!

PR #948, opened 2024-08-26
feat: vertical carousel option

ArkadiK94:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests
- [ ] Other

# Why is it needed?
- [x] users could use a vertical carousel

# Checklist:


ArkadiK94:

.changeset/metal-jobs-compete.md
```diff
@@ -0,0 +1,19 @@
+---
+'@qwik-ui/headless': minor
+---
+
+feat: add vertical carousel option
+
+- feat: add direction and maxSlideHeight props for adjusting the carousel to be vertical
+
+- chore: add directionSig to context
```
thejackshelton, line 9: for internal use we don't need to make a changeset for it.

packages/kit-headless/src/components/carousel/carousel.css
```diff
@@ -20,6 +22,7 @@
   @media (pointer: coarse) {
     [data-qui-carousel-scroller][data-draggable] {
       overflow-x: scroll;
+      overflow-y: scroll;
```
thejackshelton, line 25: maybe we can just combine this to overflow scroll?

packages/kit-headless/src/components/carousel/carousel.test.ts
```diff
@@ -486,6 +486,34 @@ test.describe('Props', () => {
       // TODO
     });
   });
+
```
thejackshelton, line 489: looks great! I wonder if we can do this with a touch device as well, since we have two separate "modes" of interaction. Mouse with JS and Touch with CSS.

apps/website/src/routes/docs/headless/carousel/index.mdx
```diff
@@ -124,6 +124,13 @@ Qwik UI supports carousels without a scroller, which can be useful for condition
 
 Remove the `<Carousel.Scroller />` component to remove the scroller.
 
+## Vertical Direction
```
thejackshelton, line 127: Let's update the CSS scroll snapping section to include the updated CSS code for vertical mode.

apps/website/src/routes/docs/headless/carousel/examples/vertical-direction.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default component$(() => {
+  useStyles$(styles);
+
+  const colors = ['red', 'green', 'blue', 'yellow', 'purple', 'orange', 'pink'];
+
+  return (
+    <Carousel.Root class="carousel-root" gap={30} direction="column" maxSlideHeight={160}>
+      <div class="carousel-buttons">
+        <Carousel.Previous>Down</Carousel.Previous>
```
thejackshelton, line 12: This might be confusing. When the Up button is pressed, it goes to the next slide, but it's scrolling below to the next slide. (Downwards)

It looks to me that the names of these should be inverted. Swiping is done upwards to get to the next slide below, but not when programmatically changing the slide.

thejackshelton:
I found a bug specifically in the mobile version, when swiping it does not update the current slide index. 

For example, swiping vertically through three slides and noticed how the previous button remains disabled. When the next button is pressed it goes to the second slide.

.changeset/metal-jobs-compete.md
ArkadiK94: Interesting, I removed this file, however, maybe, we should clarify this in the contribution guide the cases when the contributor should add changeset file [here](https://github.com/qwikifiers/qwik-ui/blob/main/CONTRIBUTING.md#6-added-a-changeset)

apps/website/src/routes/docs/headless/carousel/examples/vertical-direction.tsx
ArkadiK94: ok, I will leave it as "Next" and "Prev" I think these names can work here as well.

PR #960, opened 2024-09-13
feat: add a new switch component

JerryWu1234:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:


JerryWu1234:
image

thejackshelton:
Hey Jerry! Just went through the implementation. We're getting close!

A couple of areas of improvement:

- Clicking outside the thumb (but still on the switch itself), does not toggle the switch
- When hitting the space key, it will turn on then back off (rather than on like the enter key)
- Certain accessibility features are missing (https://www.w3.org/WAI/ARIA/apg/patterns/switch/)
- Every example seems to use a signal bind (this is usually only when the user wants to programmatically change something)

https://kobalte.dev/docs/core/components/switch/

I think this is a good source of inspiration to look at. Along with a set of resources I created here:
https://docs.google.com/document/d/18eA8WfrYtdkrRpavyoG8EmT_uzPtzt24ifOOTKk5In8/edit?tab=t.0

For example, I think it makes a lot of sense to have a new component tied to a piece of markup for the thumb, rather than having it be on a pseudo element, since that would follow the principles of composability. The idea that pieces of markup and components can blend in with each other in a straightforward way.

[Atomic Design](https://bradfrost.com/blog/post/atomic-web-design/) is a good read if you'd like to look further into it.

---

I think TDD / adding some tests would help a lot to prevent regressions and keep things robust. We need some tests before we can put the switch component in the beta stage.

JerryWu1234:
Thank you for reviewing, I will fix it soon

JerryWu1234:
- [ ] Clicking outside the thumb (but still on the switch itself), does not toggle the switch(headless done)

- [X] When hitting the space key, it will turn on then back off (rather than on like the enter key) 

- [ ] Certain accessibility features are missing (https://www.w3.org/WAI/ARIA/apg/patterns/switch/)

- [ ] Every example seems to use a signal bind (this is usually only when the user wants to programmatically change something)

JerryWu1234:
> Hey Jerry! Just went through the implementation. We're getting close!
> 
> A couple of areas of improvement:
> 
> * Clicking outside the thumb (but still on the switch itself), does not toggle the switch
> * When hitting the space key, it will turn on then back off (rather than on like the enter key)
> * Certain accessibility features are missing (https://www.w3.org/WAI/ARIA/apg/patterns/switch/)
> * Every example seems to use a signal bind (this is usually only when the user wants to programmatically change something)
> 
> https://kobalte.dev/docs/core/components/switch/
> 
> I think this is a good source of inspiration to look at. Along with a set of resources I created here: https://docs.google.com/document/d/18eA8WfrYtdkrRpavyoG8EmT_uzPtzt24ifOOTKk5In8/edit?tab=t.0
> 
> For example, I think it makes a lot of sense to have a new component tied to a piece of markup for the thumb, rather than having it be on a pseudo element, since that would follow the principles of composability. The idea that pieces of markup and components can blend in with each other in a straightforward way.
> 
> [Atomic Design](https://bradfrost.com/blog/post/atomic-web-design/) is a good read if you'd like to look further into it.
> 
> I think TDD / adding some tests would help a lot to prevent regressions and keep things robust. We need some tests before we can put the switch component in the beta stage.

I done

JerryWu1234:
it's weird, I have no idea with this error. I didn’t delete the tailwind file

packages/kit-headless/src/components/switch/switch-input.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const handleKeyPressSync$ = sync$((e: KeyboardEvent) => {
+    const keys = ['Enter', ' '];
+    if (keys.includes(e.code)) {
+      e.preventDefault();
+    }
+  });
+
+  return (
+    <div
```
thejackshelton, line 47: This breaks the rule of composability

See: https://qwik.design/contributing/composition/

I would suggest these becoming different component pieces

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        await expect(d.getTrigger()).toHaveAttribute('type', 'checkbox');
+        // role
+        await expect(d.getTrigger()).toHaveAttribute('role', 'switch');
+  })
+  test(`GIVEN a hero switch
+        WHEN toggled
+        THEN the checked property should correctly reflect the toggle state`, async ({
+    page,
+  }) => {
+    const { driver: d } = await setup(page, 'hero');
```
thejackshelton, line 42: Because this is currently on an input element, I think you want these checks to be on a different element, then checking if the attribute has data-checked

Also a different describe section for the visually hidden input

apps/website/src/routes/docs/styled/switch/examples/hero.tsx
```diff
@@ -0,0 +1,14 @@
+import { component$, useSignal } from '@builder.io/qwik';
+import { Switch } from '~/components/ui';
+
+export default component$(() => {
+  const checked = useSignal(false);
+  return (
+    <>
+      <Switch.Root class="switch" bind:checked={checked}>
+        <Switch.Label>test</Switch.Label>
```
thejackshelton, line 9: Text inside for the docs should probably be called label since it is front facing

apps/website/src/routes/docs/styled/switch/examples/hero.tsx
```diff
@@ -0,0 +1,14 @@
+import { component$, useSignal } from '@builder.io/qwik';
+import { Switch } from '~/components/ui';
+
+export default component$(() => {
+  const checked = useSignal(false);
+  return (
+    <>
+      <Switch.Root class="switch" bind:checked={checked}>
```
thejackshelton, line 8: consumer isn't doing anything with bind:checked here, internal state should be used (useBoundSignal for both)

apps/website/src/routes/docs/styled/switch/examples/disabled.tsx
```diff
@@ -0,0 +1,13 @@
+import { component$, useSignal } from '@builder.io/qwik';
+import { Switch } from '~/components/ui';
+
+export default component$(() => {
+  const checked = useSignal(false);
+  return (
+    <Switch.Root class="switch" disabled bind:checked={checked}>
```
thejackshelton, line 7: there's no reason to have a checked signal or bind:checked here if consumer is never changing the state

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    await expect(d.getTrigger()).toHaveAttribute('data-checked', 'true');
+    await expect(d.getTrigger()).toBeChecked();
+  });
+
+  test(`GIVEN a hero switch
+    WHEN clicked
+    THEN the onChange callback should be triggered`, async ({ page }) => {
+    const { driver: d } = await setup(page, 'hero');
+    await expect(d.getTriggerlaBle()).toHaveText('test0');
+    await d.getTrigger().click({force: true});
```
thejackshelton, line 55: shouldn't need force true here

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    await expect(d.getTriggerlaBle()).toHaveText('test1');
```
thejackshelton, line 56: we shouldn't be testing an onchange callback on a hero example

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    const { driver: d } = await setup(page, 'hero');
+    await d.getTrigger().focus();
+    await expect(d.getTrigger()).not.toBeChecked();
+    await d.getTrigger().press(' ');
+    await expect(d.getTrigger()).toBeChecked();
+    await d.getTrigger().press(' ');
+    await expect(d.getTrigger()).not.toBeChecked();
+  });
+
+  test(`
```
thejackshelton, line 86: This test can probably be removed, it's not really doing much

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    THEN the switch should be checked
+    `, async ({ page }) => {
+    const { driver: d } = await setup(page, 'checked');
+    await expect(d.getTrigger()).toBeChecked();
+    await expect(d.getTrigger()).toHaveAttribute('data-checked', 'true');
+    await expect(d.getTrigger()).toHaveAttribute('aria-checked', 'true');
+  });
+
+  test(`
+    GIVEN a defaultChecked switch
```
thejackshelton, line 111: GIVEN a switch that is initially checked

packages/kit-headless/src/components/switch/switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  test(`
+    GIVEN a switch without a label
+    WHEN the switch is mounted
+    THEN it should have a default label`, async ({ page }) => {
+    const { driver: d } = await setup(page, 'pure');
+    await expect(d.getTriggerlaBle()).not.toBeNull();
+  });
+
+  test(`
```
thejackshelton, line 150: Good point on this test, perhaps each of the components should have one test verifying all the main pieces have an attribute on them so that we know props is spread

thejackshelton:
I think we could merge this in as a draft state if you wanted to start consuming the current version. 

With the current feedback above a lot of changes are still needed to be prod ready. (and processes have improved since this PR)

Props to the continued effort on this @JerryWu1234 💪 . Many parts of this were not documented, and have since been documented in qwik.design

I will start writing some docs on form support as well

thejackshelton:
Hey Jerry! Here's the new docs on form handling:

https://qwik.design/contributing/forms/

thejackshelton:
Hey @JerryWu1234 let's remove this changeset since this would upgrade Qwik UI by a major version

image

JerryWu1234:
> Hey @JerryWu1234 let's remove this changeset since this would upgrade Qwik UI by a major version
> 
> image

OK

Issue #913, opened 2024-08-07
Research how to provide a better animations experience for Popover based components


cwoolum:
Because we use the native popover API as the core of some of our components, there are limitations on how animations work. Standard animations placed on a popover based component won't work because the popover immediately disappears. There is specific documentation on how to use animations with popovers.

Work needs to be done to determine if this is the best approach or if we can find another workaround. The solution should also focus on providing a great DX so users don't struggle to implement.

cwoolum:
After an analysis, here are my results

# Animations with the Popover API

## Overview


## 1. Popovers: Using `onBeforeToggle`

For popovers, the `onBeforeToggle` event provides a hook to apply animations before the popover opens or closes. This allows you to control the animation timeline, ensuring that the animation completes before any DOM changes occur.

### Example:

```javascript
export async function supportShowAnimation(popover: HTMLElement, isPolyfill: boolean) {
  popover.classList.add('popover-showing');
  popover.classList.remove('popover-closing');
  popover.dataset.open = '';
  popover.removeAttribute('data-closing');
  popover.removeAttribute('data-closed');
}

export function supportClosingAnimation(popover: HTMLElement) {
  popover.classList.remove('popover-showing');
  popover.classList.add('popover-closing');
  popover.dataset.closing = '';

  const { animationDuration, transitionDuration } = getComputedStyle(popover);

  const runAnimationEnd = () => {
    popover.classList.remove('popover-closing');
    popover.removeAttribute('data-closing');
    popover.dataset.closed = '';
  };

  const runTransitionEnd = () => {
    popover.classList.remove('popover-closing');
    popover.removeAttribute('data-closing');
    popover.dataset.closed = '';
  };

  if (animationDuration !== '0s') {
    popover.addEventListener('animationend', runAnimationEnd, { once: true });
  } else if (transitionDuration !== '0s') {
    popover.addEventListener('transitionend', runTransitionEnd, { once: true });
  } else {
    popover.classList.remove('popover-closing');
    popover.removeAttribute('data-closing');
    popover.dataset.closed = '';
  }
}
```
And the event handler
```jsx
onBeforeToggle$={[
        $(async (e) => {
          console.log('toggling')
          if (!context.panelRef?.value) return;

          if (e.newState === 'open' && context.panelRef.value) {
            await supportShowAnimation(context.panelRef.value, isPolyfillSig.value);
          }
          if (e.newState === 'closed') {
            supportClosingAnimation(context.panelRef.value);
[...]

PR #970, opened 2024-09-21
Update animations functionality to only support the native popover approach

cwoolum:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests
- [ ] Other

# Why is it needed?

This closes out #913. The existing shimming to support animations has been removed and only native popover animations will be supported. We now have full documentation along with warnings for how animations should be used with the popover API.

I also updated the modal animation documentation to reflect the current state of support.

# Checklist:


.github/workflows/test.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    name: Test NodeJS ${{ matrix.node_version }}
```
cwoolum, line 10: Adding this to make the actions page a bit more readable.

package.json
```diff
@@ -39,7 +39,7 @@
     "test.utils": "nx test utils"
   },
   "simple-git-hooks": {
-    "pre-commit": "pnpm format.staged"
+    "pre-commit": "pnpm format.staged && pnpm lint"
```
cwoolum, line 42: Added this so that linting errors are caught before creating a PR. Let me know if this was intentionally excluded.

PR #984, opened 2024-10-08
Add tree sitter to auto api

TheMcnafaha:
# What is it?

- [x] Feature / enhancement
- [x] Bug

# Why is it needed?

Current auto api cannot handle nested object-like structures. 

The solution does add two new dependencies: tree-sitter and tree-sitter-typescript

# Checklist:


maiieul:
Okay I'm merging as is, I'll make a follow up PR to cleanup a bit.

PR #985, opened 2024-10-10
refactor(auto-api.ts): chatgpt o1 cleanup

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

	•	Variable Renaming: Changed variable names like lol to capture and hacky to exportedApi for better readability.
	•	Regex Enhancements: Updated regex patterns to be platform-independent by matching both / and \ using [\\/].
	•	Edge Case Handling: Added checks to prevent accessing undefined variables when arrays may be empty.
	•	Error Logging: Included error logging in the catch block of writeToDocs to output any errors during file writing.
	•	Path Handling: Used path.resolve to handle file paths in a platform-independent way.
	•	Comment Trimming: Trimmed comments to remove unnecessary whitespace.

# Checklist:

### DIFF
--- a/apps/website/auto-api.ts
+++ b/apps/website/auto-api.ts
@@ -67,7 +67,27 @@ type ParsedProps = {
 
 /**
  * Note: For this code to run, you need to prefix the type with 'Public' (e.g., 'PublicMyType') in your TypeScript files
- * and save the file. This convention helps the parser identify and process the public types correctly.
+ *
+ *  * e.g:
+ *
+ * ```tsx
+ * type PublicModalRootProps = {
+ *     //blablabla
+ *     onShow$?: QRL<() => void>;
+ *     //blablabla
+ *     onClose$?: QRL<() => void>;
+ *     //blablabla
+ *     'bind:show'?: Signal<boolean>;
+ *     //blablabla
+ *     closeOnBackdropClick?: boolean;
+ *     //blablabla
+ *     alert?: boolean;
+ * };
+ * ```
+ * This convention helps the parser identify and process the public types correctly.
+ *
+ * Now when you save the corresponding .mdx file, the API will be updated accordingly.
+ *
  **/
 
 function parseSingleComponentFromDir(
@@ -104,9 +124,9 @@ function parseSingleComponentFromDir(
       }
       if (capture.name === 'type') {
         parsedProps.type = capture.node.text;
-        const lastKey = last ? Object.keys(last)[0] : '';
-        if (subComponentName === lastKey) {
-          last![lastKey].push(parsedProps);
+        const topKey = last ? Object.keys(last)[0] : '';
+        if (subComponentName === topKey) {
+          last[topKey].push(parsedProps);
         } else {
           parsed.push({ [subComponentName]: [parsedProps] });
         }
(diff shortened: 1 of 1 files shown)


apps/website/auto-api.ts
```diff
@@ -63,93 +64,99 @@ type ParsedProps = {
...
-      return;
+      console.error('Error writing API file:', err);
     }
   }
 }
+
 function loopOnAllChildFiles(filePath: string) {
-  const childComponentRegex = /\/([\w-]*).tsx$/.exec(filePath);
-  if (childComponentRegex === null) {
+  const childComponentMatch = /[\\/](\w[\w-]*)\.tsx$/.exec(filePath);
```
TheMcnafaha, line 144: good catch on escaping the dot. I'm still curious why the AI insist on having windows support  though

apps/website/auto-api.ts
```diff
@@ -63,93 +64,99 @@ type ParsedProps = {
...
+  if (!childComponentMatch) {
     return;
   }
-  const parentDir = filePath.replace(childComponentRegex[0], '');
-  const componentRegex = /\/(\w*)$/.exec(parentDir);
-  if (!fs.existsSync(parentDir) || componentRegex == null) {
+  const parentDir = filePath.slice(0, filePath.lastIndexOf(childComponentMatch[0]));
```
TheMcnafaha, line 148: I'm not sure if this is more readable that what it replaces. 

```js
const parentDir=filePath.replace(childComponentRegex[0], '')

const parentDir = filePath.slice(0, filePath.lastIndexOf(childComponentMatch[0]));
```
They both have the regex[0] syntax, but one uses the slice trick and the other just uses replace. Maybe slice is more popular than replace? Though I think the word "replace" tells you everything you need to know.

apps/website/auto-api.ts
```diff
@@ -63,93 +64,99 @@ type ParsedProps = {
...
   }
-  const componentName = componentRegex[1];
+  const componentName = componentMatch[1];
   const allParts: SubComponents = [];
   const store: ComponentParts = { [componentName]: allParts };
+
   fs.readdirSync(parentDir).forEach((fileName) => {
-    if (/tsx$/.test(fileName)) {
-      const fullPath = parentDir + '/' + fileName;
+    if (/\.tsx$/.test(fileName)) {
```
TheMcnafaha, line 158: Another good catch. While it would take some adventurous file naming, the previous regex could give false positives

TheMcnafaha (commented):
I think gpt knows node better than me lol. It did make some interesting decisions regarding some of the regex, but if you're short on time I think you can just merge the file as it. It's a net improvement anyways.

apps/website/auto-api.ts
```diff
@@ -63,93 +64,99 @@ type ParsedProps = {
...
   if (fullPath.includes('kit-headless')) {
-    const relDocPath = `../website/src/routes//docs/headless/${componentName}`;
+    const relDocPath = `../website/src/routes/docs/headless/${componentName}`;
     const fullDocPath = resolve(__dirname, relDocPath);
-    const dirPath = fullDocPath.concat('/auto-api');
+    const dirPath = resolve(fullDocPath, 'auto-api');
 
     if (!fs.existsSync(dirPath)) {
-      fs.mkdirSync(dirPath);
+      fs.mkdirSync(dirPath, { recursive: true });
```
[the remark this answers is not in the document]
maiieul: Ah I guess it's not needed as long as the generated fire isn't nested.

maiieul:
Took your comments into account. Thanks for the review 🙇‍♂️ . I think the regex is good for windows. But if it's problematic feel free to re-change in another PR :)

PR #997, opened 2024-10-16
fix: useComputed$ double read

thejackshelton:
In v2 combobox and dropdown break because of a computed that does a signal read of another computed.

This works as a workaround (but is also more efficient) to the bug in v2.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

# Checklist:


thejackshelton:
fixed in v2 so this workaround is no longer needed

PR #1046, opened 2025-02-12
fix: recursive alias imports

maiieul:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

cc @thejackshelton we need to avoid importing from e.g. @qwik-ui/headless inside the headless package itself, otherwise the optimizer adds unrelated imports to a component when building the lib, which leads to over-prefetching to consumers of that lib.

image
Here the toggle-group-item imports a lot of components from Accordion and Dropdown (because it's an MRE and there's only Accordion, Dropdown, Toggle and ToggleGroup). This will lead to over-prefetching.

# Checklist:


thejackshelton:
Thanks for catching that! Yes this is not intended.

PR #1110, opened 2025-07-02
Fix modal.close not being of type button

Caesarovich:
# What is it?

- [x] Fix

# Why is it needed?

This is needed because it's a custom button and it is more semantically correct. This also avoid problems such as the browse coercing it into a "submit" button if it is placed within a form. (Which caused my forms to submit when closing the modal).

# Checklist:


Caesarovich:

PR #1167, opened 2025-11-20
fix: use type="button" for trigger components

fintanmartin:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests
- [ ] Other

# Why is it needed?

\<button\> elements default to type="submit", which submit when clicked if used inside a form.
This leads to an annoying issue where using components like Modal inside of forms will submit the form when you attempt to interact with them.

This small patch fixes that by specifying type="button" in all of the .Trigger components.

# Checklist:


fintanmartin:

fintanmartin:
Didn't realise that the components exposed all the props of the underlying button including the type, probably not needed.

PR #1175, opened 2026-04-17
fix(modal): prevent onClose$ from firing on initial mount when show is false

FHachez:
## Summary

- `onClose$` was being called on the initial `useTask$` run whenever the modal started in a closed state (`bind:show = false`), because the task unconditionally fell into the `else` branch on first execution.
- Added a `wasOpenSig` flag that is set to `true` the first time the modal opens. `onClose$` is now only invoked after a real open → closed transition.
- `onShow$` behavior is unchanged.

## Root cause

In Qwik, `useTask$` runs once on mount to register tracking, then re-runs on signal changes. On the first run with `showSig.value === false`, the previous code hit the `else` branch and called `onClose$` — even though the modal had never been open.

### DIFF
--- a/.changeset/floppy-donuts-begin.md
+++ b/.changeset/floppy-donuts-begin.md
@@ -1,0 +1,5 @@
+---
+'@qwik-ui/headless': patch
+---
+
+fix: onClose$ now doesn't fire if the show prop is set to false by default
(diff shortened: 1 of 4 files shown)


FHachez:

